PR #18, opened 2014-12-05
Implement worker-to-worker shard copy UDF

jasonmp85:
The master node already has a UDF to repair an existing placement, but it requires a corresponding UDF on the worker side to actually copy the data between workers. This PR implements that UDF.

**Code Review Tasks**

  - [x] Remove `copy_shard` files and consolidate code in `repair_shards` files
  - [x] Improve error message style
  - [x] Reorder `CopyDataFromFinalizedPlacement` arguments
  - [x] Consider use of `ExecuteRemoteCommand`
  - [x] Investigate need for `heap_open`/`heap_open`
  - [x] Check whether `ExecMaterializeSlot` creates a copy
  - [x] Change `ExecuteRemoteCommand` to permit `PGRES_TUPLES_OK`
  - [x] Call `ExecuteRemoteCommandList` instead of handling manually
  - [x] Delay `relationKind` assignment until right before `if` block


copy_shard.c
```diff
@@ -0,0 +1,139 @@
...
+#include "utils/tuplestore.h"
+
+
+/* local function forward declarations */
+static bool CopyDataFromTupleStoreToRelation(Tuplestorestate *tupleStore,
+											 Relation relation);
+
+
+/* declarations for dynamic loading */
+PG_FUNCTION_INFO_V1(worker_copy_shard_placement);
```
ozgune: What if we kept this function as part of repair_shards.h/.c for now?

I realize we could conceptually use this UDF in places other than shard repairs. For now, our implementation in this file is 100-200 lines. Therefore, I have a slight preference for keeping this in repair_shards, and consider splitting this once we have 3 separate callers to this function. Any strong objections?

copy_shard.c
```diff
@@ -0,0 +1,139 @@
...
+Datum
+worker_copy_shard_placement(PG_FUNCTION_ARGS)
+{
+	text *shardRelationNameText = PG_GETARG_TEXT_P(0);
+	text *nodeNameText = PG_GETARG_TEXT_P(1);
+	int32 nodePort = PG_GETARG_INT32(2);
+	Oid shardRelationId = ResolveRelationId(shardRelationNameText);
+	char *shardRelationName = text_to_cstring(shardRelationNameText);
+	char *nodeName = text_to_cstring(nodeNameText);
+
```
ozgune: Two minor notes.

1. We have an extra newline between these blocks. Could we remove it?
2. To keep the distance low between variable declarations and definitions, what do you think of declaring ``placement``, ``task``, and ``selectAllQuery`` as NULL, and then defining them just before their use below?

copy_shard.c
```diff
@@ -0,0 +1,139 @@
...
+	TupleDesc tupleDescriptor = RelationGetDescr(relation);
+	TupleTableSlot *tupleTableSlot = MakeSingleTupleTableSlot(tupleDescriptor);
+	bool copySuccessful = false;
+
+	for (;;)
+	{
+		HeapTuple tupleToInsert = NULL;
+		Oid insertedOid = InvalidOid;
+		bool nextTuple = tuplestore_gettupleslot(tupleStore, true, false, tupleTableSlot);
+		if (!nextTuple)
```
ozgune: Two questions.

1. Do we need the ``copySuccessful`` flag? The functions in this ``for (;;)`` seem to report their errors by calling ``ereport(ERROR, ...)``` Does any one of them report their errors through their return values?
2. I'll chat with Sumedh on ``TupleStoreToTable`` function. I'm looking to understand if there's a way in which we can refactor that function and use it here.

repair_shards.c
```diff
@@ -186,11 +179,13 @@ RecreateTableDDLCommandList(Oid relationId, int64 shardId)
 	/* build appropriate DROP command based on relation kind */
 	if (relationKind == RELKIND_RELATION)
 	{
-		appendStringInfo(extendedDropCommand, DROP_REGULAR_TABLE_COMMAND, shardName);
+		appendStringInfo(extendedDropCommand, DROP_REGULAR_TABLE_COMMAND,
+						 quote_identifier(shardName));
```
ozgune: We quote the ``shardName`` just before this ``if`` block. Do we need to call ``quote_identifier`` again?

repair_shards.c
```diff
@@ -98,9 +102,12 @@ master_copy_shard_placement(PG_FUNCTION_ARGS)
 	}
 
 	HOLD_INTERRUPTS();
-
-	/* TODO: implement row copying functionality */
-	ereport(ERROR, (errmsg("shard placement repair not fully implemented")));
+	dataCopied = CopyDataFromFinalizedPlacement(sourcePlacement, targetPlacement,
```
ozgune: Two quick notes.

1. We deleted a newline between ``HOLD_INTERRUPTS`` and the ``dataCopied``. Could we add back the newline?
2. Could we make this error message and the one just before it similar? For example, "failed to copy placement data" -> "could not copy placement data" This wording also plays nicer with PostgreSQL's conventions on wording error messages.

repair_shards.c
```diff
@@ -209,3 +204,67 @@ RecreateTableDDLCommandList(Oid relationId, int64 shardId)
...
+	char *relationName = get_rel_name(distributedTableId);
+	char relationKind = get_rel_relkind(distributedTableId);
+	const char *shardName = NULL;
+	char *nodeName = placementToRepair->nodeName;
+	int32 nodePort = placementToRepair->nodePort;
+	StringInfo copyRelationQuery = makeStringInfo();
+
+	PGconn *connection = NULL;
+	PGresult *result = NULL;
+	char *copySuccessfulString = NULL;
```
ozgune: We don't use ``copySuccessfulString`` and ``responseParsedSuccessfully`` variables in this function. Could we remove them?

repair_shards.c
```diff
@@ -69,6 +72,7 @@ master_copy_shard_placement(PG_FUNCTION_ARGS)
 	ShardPlacement *targetPlacement = NULL;
 	List *ddlCommandList = NIL;
 	bool recreated = false;
+	bool dataCopied = false;
```
ozgune: We have the declaration ``ShardPlacement *sourcePlacement PG_USED_FOR_ASSERTS_ONLY = NULL;``.

Could we remove the part on ``PG_USED_FOR_ASSERTS_ONLY`` from the declaration?

repair_shards.c
```diff
@@ -209,3 +204,67 @@ RecreateTableDDLCommandList(Oid relationId, int64 shardId)
...
+
+/*
+ * CopyDataFromFinalizedPlacement connects to an unhealthy placement and
+ * directs it to copy the specified shard from a certain healthy placement.
+ * This function assumes that the unhealthy placement already has a schema
+ * in place to receive rows from the healthy placement. This function returns
+ * a boolean indicating success or failure.
+ */
+static bool
+CopyDataFromFinalizedPlacement(ShardPlacement *placementToRepair,
```
ozgune: Two quick notes on the function declaration.

1. Do we have a discrepancy between the order of ``ShardPlacement`` arguments? The caller first uses the source placement (healthy placement) and the target placement (placementToRepair). I think we declare those arguments in the inverse order in here, but I'm not sure.
2. To be consistent in the order of arguments between ``RecreateTableDDLCommandList`` and this function, what do you think of the following argument ordering? ``Oid distributedTableId, int64 shardId, ShardPlacement *sourcePlacement, ShardPlacement *targetPlacement``? (This way, we'd follow the consistency advice in CC2 7.5)

repair_shards.c
```diff
@@ -209,3 +204,67 @@ RecreateTableDDLCommandList(Oid relationId, int64 shardId)
...
+	{
+		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						errmsg("cannot repair shard"),
+						errdetail("Repairing shards backed by foreign tables is "
+								  "currently unsupported.")));
+	}
+
+	AppendShardIdToName(&relationName, shardId);
+	shardName = quote_identifier(relationName);
+
```
ozgune: What if we made ``ExecuteRemoteCommand`` public, and directly called it here?

If we call the ``ExecuteRemoteCommand`` function, what do you think of inlining this logic in ``master_copy_shard_placement``?

repair_shards.h
```diff
@@ -20,11 +20,11 @@
 /* templates for SQL commands used during shard placement repair */
 #define DROP_REGULAR_TABLE_COMMAND "DROP TABLE IF EXISTS %s"
 #define DROP_FOREIGN_TABLE_COMMAND "DROP FOREIGN TABLE IF EXISTS %s"
+#define COPY_RELATION_QUERY "SELECT worker_copy_shard_placement(%s, %s, %d)"
```
ozgune: A few ``#define`` name alternatives to make the name similar to Citus (we use TABLE_FETCH_COMMAND in Citus). What do you think about the following alternatives?

1. COPY_PLACEMENT_COMMAND or COPY_SHARD_PLACEMENT_COMMAND
2. COPY_PLACEMENT_QUERY or COPY_SHARD_PLACEMENT_QUERY

repair_shards.c
```diff
@@ -209,3 +204,67 @@ RecreateTableDDLCommandList(Oid relationId, int64 shardId)
...
+		ereport(WARNING, (errmsg("could not connect to %s:%d", nodeName, nodePort)));
+
+		return false;
+	}
+
+	appendStringInfo(copyRelationQuery, COPY_RELATION_QUERY,
+					 quote_literal_cstr(shardName),
+					 quote_literal_cstr(healthyPlacement->nodeName),
+					 healthyPlacement->nodePort);
+
```
ozgune: Could we remove one of the newlines here?

copy_shard.c
```diff
@@ -0,0 +1,139 @@
...
+		{
+			copySuccessful = true;
+			break;
+		}
+
+		tupleToInsert = ExecMaterializeSlot(tupleTableSlot);
+
+		insertedOid = simple_heap_insert(relation, tupleToInsert);
```
ozgune: Minor. Do we need the ``insertedOid`` variable?

copy_shard.c
```diff
@@ -0,0 +1,139 @@
...
+		HeapTuple tupleToInsert = NULL;
+		Oid insertedOid = InvalidOid;
+		bool nextTuple = tuplestore_gettupleslot(tupleStore, true, false, tupleTableSlot);
+		if (!nextTuple)
+		{
+			copySuccessful = true;
+			break;
+		}
+
+		tupleToInsert = ExecMaterializeSlot(tupleTableSlot);
```
ozgune: Does the ``ExecMaterializeSlot`` create a copy for the ``tupleToInsert``? (don't know, just asking)  _If so_, do we need to free this copy?

copy_shard.c
jasonmp85: Agreed. I'll move it over. I had it in a separate file back when I thought we might build separate packages for the master and worker nodes so didn't change it but was going to do so if you brought it up.

copy_shard.c
jasonmp85: 1. Sure, didn't notice that before committing; sorry
2. I've reworked this section to have a bit better parallel structure based on your suggestion.

copy_shard.c
jasonmp85: I should have caught the `insertedOid` thing myself with `scan-build`, sorry:

screen shot 2015-03-04 at 1 18 38 pm

copy_shard.c
jasonmp85: I don't know where that return value came from. I'm guessing I started with it and slowly removed everything it was needed for. Removing and switching function to `void`.

repair_shards.c
jasonmp85: Nope! Fixed.

repair_shards.c
jasonmp85: 1. Yup.
2. I'll look over all error messages shortly and make them fit with conventions.

repair_shards.c
jasonmp85: Yes. I have _no_ idea why analysis didn't catch this.

repair_shards.c
jasonmp85: Good catch.

repair_shards.h
jasonmp85: I guess I was distinguishing between commands (that can modify) and queries (that can't). But that's a false distinction… this `SELECT` (a "query") is making modifications by calling a funtion (so it's a "command"). I'll go with `COPY_SHARD_PLACEMENT_COMMAND`

repair_shards.c
jasonmp85: Yes.

jasonmp85:
I added a task list to this review. I've (locally) done most of the small changes but will check off each of the larger ones as they're addressed.

repair_shards.c
jasonmp85: Two things:

  * I'm using a `SELECT` to call a function here, so `ExecuteRemoteCommand` doesn't work (it requires commands that return an OK status, not tuples)
  * Even if we could reduce the lines of code by calling that, this function does enough that inlining it would be very distracting. I think it needs to stay as-is.

copy_shard.c
jasonmp85: On the second question: `ExecMaterializeSlot` _does_ allocate memory and it first appears that this means we need to call `heap_freetuple` to avoid memory problems here; however, it also sets `tupleTableSlot->tts_shouldFree` to `true`, which causes `ExecClearTuple` to call `pfree` on `tupleTableSlot->tts_tuple`, i.e. the tuple.

So in short, there is no memory problem here. We load a (non-copy) tuple into the slot, create a copy (the heap tuple) to do the insert, then free it (because the flag set on the slot causes the clear call to free the associated tuple).

copy_shard.c
jasonmp85: `ExecMaterializeSlot` _does_ allocate memory and it first appears that this means we need to call `heap_freetuple` to avoid memory problems here; however, it also sets `tupleTableSlot->tts_shouldFree` to `true`, which causes `ExecClearTuple` to call `pfree` on `tupleTableSlot->tts_tuple`, i.e. the tuple.

So in short, there is no memory problem here. We load a (non-copy) tuple into the slot, create a copy (the heap tuple) to do the insert, then free it (because the flag set on the slot causes the clear call to free the associated tuple).

copy_shard.c
jasonmp85: The call to `simple_heap_insert` _is_ wrapped with `heap_open` and `heap_close`… the relation is already open when we enter this function (the caller opens and closes it).

I'm going to update the function comments to make it clear that it expects the relation to be open.

repair_shards.c
jasonmp85: 1. Fun fact: my unit tests pass even if I switch the order of arguments here. Side effect of hacking together a shard repair test using a single PostgreSQL instance. Chalk it up as another vote for "need to make better integration testing rig, eventually". Good catch, that was a bug that manual testing would have caught (which I did a few months ago but had refactored since, trusting my test).
2. Yes, changing.

jasonmp85:
I've addressed all of your feedback. Next round!

repair_shards.c
```diff
@@ -115,14 +141,54 @@ master_copy_shard_placement(PG_FUNCTION_ARGS)
...
+	text *shardRelationNameText = PG_GETARG_TEXT_P(0);
+	text *nodeNameText = PG_GETARG_TEXT_P(1);
+	int32 nodePort = PG_GETARG_INT32(2);
+	char *shardRelationName = text_to_cstring(shardRelationNameText);
+	char *nodeName = text_to_cstring(nodeNameText);
+	bool fetchSuccessful = false;
+
+	Oid shardRelationId = ResolveRelationId(shardRelationNameText);
+	Relation shardTable = heap_open(shardRelationId, RowExclusiveLock);
+	TupleDesc tupleDescriptor = RelationGetDescr(shardTable);
```
ozgune: What if we called ``heap_close`` after getting the tuple descriptor?

That way, we could independently call ``heap_open`` / ``heap_close`` in ``CopyDataFromTupleStoreToRelation`` and encapsulate all heap logic inside that function.

One question this brings is what happens if the table's schema changes from the time we open it here to the copy function. (If that's the case, the data we fetched from the remote node could also be in an incorrect format. Maybe the best way to ensure that is through the exclusive shardId lock on the master node.)

repair_shards.c
```diff
@@ -115,14 +141,54 @@ master_copy_shard_placement(PG_FUNCTION_ARGS)
...
+
+	fetchSuccessful = ExecuteTaskAndStoreResults(task, tupleDescriptor, tupleStore);
+	if (!fetchSuccessful)
+	{
+		ereport(ERROR, (errmsg("could not store shard rows from healthy placement"),
+						errhint("Consult recent messages in the server logs for "
+								"details.")));
+	}
+
+	CopyDataFromTupleStoreToRelation(tupleStore, shardTable);
```
ozgune: Minor naming question.

CopyDataFromTupleStoreToRelation --> TupleStoreToRelation ? The shorter name is more similar to TupleStoreToTable that we use in pg_shard.c.

repair_shards.c
ozgune: Got it. Let's keep this logic in a separate function. Two minor notes:

1. In other places in the code, we assumed that GetConnection() would emit the proper warning message, and simply skipped logging a warning.
2. We partially modeled ``ExecuteRemoteCommand`` on the function in CitusDB (stage.c). In Citus, ``ExecuteRemoteCommand`` both checks against ``PGRES_COMMAND_OK`` and ``PGRES_TUPLES_OK``. If it helps, we could keep a similar structure in pg_shard. What do you think?

repair_shards.c
```diff
@@ -209,3 +276,99 @@ RecreateTableDDLCommandList(Oid relationId, int64 shardId)
...
+	char relationKind = get_rel_relkind(distributedTableId);
+	const char *shardName = NULL;
+	char *nodeName = placementToRepair->nodeName;
+	int32 nodePort = placementToRepair->nodePort;
+	StringInfo copyRelationQuery = makeStringInfo();
+
+	PGconn *connection = NULL;
+	PGresult *result = NULL;
+
+	if (relationKind == RELKIND_FOREIGN_TABLE)
```
ozgune: Pretty minor. What if we moved /grouped the ``char relationKind = get_rel_relkind(..)`` declaration with this ``if`` block?

repair_shards.c
jasonmp85: 1. That sounds correct. I'll just change it to return if the connection is `NULL`.
2. The `ExecuteRemoteCommand` in `stage.c` looks most analogous to what we have here, but it returns the actual `PGresult`. Rather than rework the rest of our code to expect that, I'll keep it returning `bool` but add `PGRES_TUPLES_OK` to the check. So the contract will be you can send any query, so long as all you care about is whether it succeeded (and not the actual result). Then I'll switch to using it here.

repair_shards.c
jasonmp85: On that question: aren't there MVCC snapshots for catalog data (such as `pg_class`, etc.) that would prevent that kind of race condition (i.e. provide a consistent view of our schema throughout the transaction)? If so, shouldn't we use one of those in this case?

I'll do what you suggest with the encapsulation.

repair_shards.c
jasonmp85: I agree somewhat, but there's also the method `CopyDataFromFinalizedPlacement` in this same file, so `CopyDataFromTupleStoreToRelation` is more analogous to it. I'll think about what I want to do (unsure right now).

repair_shards.c
jasonmp85: No objections. Sure.

repair_shards.c
jasonmp85: Actually, I just realized the only `extern` function at the moment is `ExecuteRemoteCommandList`. Since it also sets up the connection, I can call it to do what I want here. I'll do that; will save on LoC and repetition.

repair_shards.c
jasonmp85: Hm, actually on second thought it would make the enclosing function less readable, and your schema concern is a real one. Keeping the relation open the whole time we need it sidesteps all of that; resorting to an external lock to accomplish something we could just do by getting the proper one on the relation feels hackier than just using the relation directly.

repair_shards.c
jasonmp85: Yeah, these two functions are `static` within `repair_shards`, so I like them being consistent within each other internal to the file. Leaving as-is.

PR #35, opened 2014-12-15
Feature issue#12

### LINKED ISSUE #12: Shards backed by foreign tables not marked as such
_From @samay-sharma on November 21, 2014 15:12_

Currently, the storage type is always set as `‘t’` in `pgs_distribution_metadata.shard`. We also have a `SHARD_STORAGE_FOREIGN` in distribution_metadata.h. I think we should use that for foreign tables.

Also, this does not seem to hamper anything. I tried with cstore_fdw and `INSERT` commands for foreign tables are routed to the correct worker shards. The worker node commands fail as `INSERT` is not yet implemented for cstore_fdw but pg_shard logic on the master seems to work correctly.

onderkalaci:
Shards backed by foreign tables marked correctly in the metadata.

Tested with:

1. Postgres 9.4
    * foreign table - `SELECT`/`INSERT`
    * normal table (`RELKIND_RELATION`)  - `SELECT`/`INSERT`
2. Citus 3.0.1
    * foreign table - `SELECT` problematic - sent mail to @jasonmp85. But I don't think it is related to this fix, because when I test with undoing the fix, we still have the problem.
    * foreign table - `INSERT` OK
    * normal table (`RELKIND_RELATION`)  - `select`/`insert`

closes #12

Review tasks:

  - [x] Make call to `get_rel_relkind` during declaration of `relationKind`
  - [x] Move declaration of `shardStorageType` to immediately after `relationKind` declaration
  - [x] Add tests for this functionality (distribute a foreign table and verify the `'f'` using a `SELECT`)


onderkalaci:
@jasonmp85 Since it is the first time I wrote regression tests, it would be nice if you can check it in detail. 
I had to change expected/modifications.out.tmpl", is that required or useless? Is there a way not to do that?

jasonmp85:
If you look into the `Makefile`, you'll see that the `tmpl` files are used to generate the `sql` and `out` files for certain tests. This is because certain tests' output will contain the port number of the PostgreSQL instance being used and since `make installcheck` just checks against the current installation, we have no way of predicting what that will be. So if a `tmpl` file exists, it should be what you modify; otherwise, just modify the `sql` and `out` files as necessary.

sql/create_shards.sql.tmpl
```diff
@@ -79,3 +79,32 @@ SELECT sort_names('sumedh', 'jason', 'ozgun');
...
+
+SELECT master_create_distributed_table('foreign_table_to_distribute', 'id');
+SELECT master_create_worker_shards('foreign_table_to_distribute', 16, 1);
+
+\set VERBOSITY default
+SELECT storage, min_value, max_value FROM pgs_distribution_metadata.shard
+	WHERE relation_id = 'foreign_table_to_distribute'::regclass
+	ORDER BY (min_value COLLATE "C") ASC;
+
+-- cleanup foreign table, related shards and shard placements
```
jasonmp85: I had been adding e.g. `DROP TABLE` statements to tests as necessary to ensure each file ran on a mostly clean database, but hadn't thought about the things in `pgs_distribution_metadata`. I'm not sure whether we want to clean it out between each test or not, but either way we'll want some unification across our tests. You don't need to change anything here, I just wanted to make a note to myself that we need a little bit of cleanup in this regard.

jasonmp85:
This looks good! Merging.

PR #38, opened 2014-12-19
Misleading error for incompletely created tables

### LINKED ISSUE #9: Misleading error for incompletely created tables
_From @samay-sharma on November 19, 2014 15:38_

I get this error when I run a select query after running `master_create_distributed_table()` but before running `master_create_worker_shards()`. The same query works after running `master_create_worker_shards()`.

Steps to reproduce:

``` sql
CREATE TABLE customer_reviews
(
    customer_id INTEGER,
    review_date DATE,
    review_rating INTEGER,
    review_votes INTEGER,
    review_helpful_votes INTEGER,
    product_id CHAR(10),
    product_title TEXT,
    product_sales_rank BIGINT,
    product_group TEXT,
    product_category TEXT,
    product_subcategory TEXT,
    similar_product_ids CHAR(10)[]
);
# CREATE TABLE

SELECT master_create_distributed_table('customer_reviews', 'customer_id');
#  master_create_distributed_table
# ---------------------------------
#  
# (1 row)

select count(*) from customer_reviews ;
# ERROR:  cannot execute select over multiple shards
```

I had expected to get 0 as the result. Even if we error out, the error message does not seem very apt.

onderkalaci:
Before this fix, if you try to execute any query on tables which are distributed with `master_create_distributed_table` but no shards are created yet for the table (i.e. `master_create_worker_shards` not called), you get unclear error message. This fix catches that case implicitly and more meaningful message is shown.

fixes #9

Review tasks:
- [x] Add more specific error code (`ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE`)
- [x] Change message, detail, and hint per my suggestions
- [x] Move `NIL` check into `DistributedQueryShardList` itself
- [x] Update `DistributedQueryShardList`'s function comment to reflect its new "return non-empty list or error" behavior
- [x] Ensure tests all pass

### DIFF
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -549,14 +549,13 @@ DistributedQueryShardList(Query *query)
 	/* error out if no shards exists for the table */
 	if (shardIntervalList == NIL)
 	{
-		char *relName = get_rel_name(distributedTableId);
+		char *relationName = get_rel_name(distributedTableId);
 
 		ereport(ERROR, (errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
 						errmsg("could not find any shards for query"),
-						errdetail("No shards exist for distributed table"
-								  " \"%s\".", relName),
-						errhint("Run master_create_worker_shards to "
-								"create shards "
+						errdetail("No shards exist for distributed table \"%s\".",
+								  relationName),
+						errhint("Run master_create_worker_shards to create shards "
 								"and try again.")));
 	}
 


onderkalaci:
With this implementation, we couldn't specify the name of the relation for which shards are not created in the ereport (Well, sure we can specify but code gets longer and become complicated).
    Thus, I had to use "the distributed table" phrase instead of its name in the ereport. Also, I think we are sure that there is only a single table, because we error out in the "ErrorIfQueryNotSupported" function if there are more than one relation involved in the query.

pg_shard.c
```diff
@@ -226,8 +226,18 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
...
 
-		/* compute the list of shards this query needs to access */
+		/*
+		 * Compute the list of shards this query needs to access.
+		 * Error out if there are no existing shards for the table.
+		 */
 		queryShardList = DistributedQueryShardList(distributedQuery);
+		if (queryShardList == NIL)
+		{
+			ereport(ERROR, (errmsg("cannot plan SELECT query"),
```
jasonmp85: This message could be generated by an `INSERT`, `UPDATE`, or `DELETE` as well…

jasonmp85: Also, the style guide says this [should be in past tense](http://www.postgresql.org/docs/9.4/static/error-style-guide.html#AEN104799): _could not plan query_.

pg_shard.c
```diff
@@ -226,8 +226,18 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
...
+					errdetail("There are no existing shards for the distributed table."),
```
jasonmp85: The indentation here is off (should be aligned with parenthesis right before `errmsg`, not those before `ERROR`).

pg_shard.c
jasonmp85: I think we can use `ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE` here: we have an object (the distributed table) which hasn't been initialized yet (by creating shards). This code is used in PostgreSQL for other analogous situations, such as not calling `nextval` for a sequence before asking it what its current value is (`grep` for it).

jasonmp85:
I was envisioning that the check would actually live in `DistributedQueryShardList` rather than after the call to it. Right now that method always returns a list, which can be the empty list (`NIL`). I'm proposing we change it to always return a non-empty list or error.

Getting the relation name would also be easy because within that method the `distributedTableId` is in scope. It would just need a block like this:

```c
List *prunedShardList = PruneShardList(distributedTableId, restrictClauseList, shardIntervalList);

if (prunedShardList == NIL)
{
    char *relname = get_rel_name(distributedTableId);

    ereport(ERROR, (errmsg("... %s ...", relname)));
}

return prunedShardList;
```

As for the parts of the error message:

  * `errcode` — we should use `ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE`
  * `errmsg` — should read _could not find any shards for query_
  * `errdetail` — should read _No shards exist for distributed table "%s"._
  * `errhint` — Don't need the quotation marks around the method name: _Run master_create_worker_shards to create shards and try again._

My suggestions for the above are based on my reading of [the style guide](http://www.postgresql.org/docs/9.4/static/error-style-guide.html). I think we should avoid using quotes unless they contain dynamic strings that could contain space-separated words. If we know ahead of time that a function name looks like a function name, the quotes are "unnecessary" in the words of the style guide. Additionally, we should say _could not_ when the user can take an action to fix something and _cannot_ if it will remain impossible forever.

jasonmp85:
Alright, I'm going to add some checklist items to get this wrapped up. I'm being strict about error messages, but we've been lax about doing them properly, saying we'd clean them up later. So I just want to make sure we are being better about them going forward.

After the checklist items are complete I can probably merge this Thursday night or Friday (US time).

jasonmp85:
Added checklist up top! Push up some changes to address those things and we'll have a :shipit:!

onderkalaci:
Below are my comments on the changes:

* In your second comment, you asked me to check *prunedShardList*, however, I think checking *shardIntervalList* is more convinient, as you had already suggested before. 

* With this changes *DistributedQueryShardList* became more complex compared to the previous implementation. But, I think we need to change the code such that we delay calls to *QueryRestrictList* and *PruneShardList* until we check existence of shards.

* I totally agree with you for the style and content of error messages. Even I checked them more than once, I still have mistakes. In the future changes, I'll pay more attention to the messages. 
Especially for the "could not vs cannot" discussion, I missed the senctence in the postgres documentation, which is "perhaps after fixing some problem". Certainly, for our case the user can fix the problem.

* For the error codes, I think I may need your help in the future. There are quite a lot of different error codes, and it is difficult for me to find and decide the correct one yet.

* Lastly, since we updated the function comment of *DistributedQueryShardList*, I need to be sure that *PruneShardList* never returns NIL when a non-NIL *shardIntervalList* is provided.
I think this is the case since we are making hash pruning.  Am I missing something on that?

pg_shard.c
```diff
@@ -530,17 +533,36 @@ ExtractRangeTableEntryWalker(Node *node, List **rangeTableList)
...
-	List *restrictClauseList = QueryRestrictList(query);
+	Oid distributedTableId = ExtractFirstDistributedTableId(query);
 	List *shardIntervalList = LookupShardIntervalList(distributedTableId);
-	List *prunedShardList = PruneShardList(distributedTableId, restrictClauseList,
-										   shardIntervalList);
+
+	/* error out if no shards exists for the table */
+	if (shardIntervalList == NIL)
+	{
+		char *relName = get_rel_name(distributedTableId);
```
jasonmp85: We use `relationName` for relation names in our code (PostgreSQL uses `relname`, no uppercase, but we prefer the full word).

pg_shard.c
```diff
@@ -530,17 +533,36 @@ ExtractRangeTableEntryWalker(Node *node, List **rangeTableList)
...
+
+		ereport(ERROR, (errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+						errmsg("could not find any shards for query"),
+						errdetail("No shards exist for distributed table"
```
jasonmp85: We wrap code at 90 characters, so this full format string can fit on one line here (it's OK that the second argument is orphaned on the next line).

pg_shard.c
```diff
@@ -530,17 +533,36 @@ ExtractRangeTableEntryWalker(Node *node, List **rangeTableList)
...
+								  " \"%s\".", relName),
+						errhint("Run master_create_worker_shards to "
```
jasonmp85: Same story here. This should fit on two lines and is weird split across three like this.

jasonmp85:
Two small issues. Fix them and :shipit:. You can merge it yourself with the _Merge pull request_ button or do it in the `git` CLI.

Before shipping:

  * Change `relName` to `relationName`
  * Rewrap the detail and hint strings to 90 columns

Issue #47, opened 2014-12-27
VOLATILE functions accepted in non-partition key qual. clauses


jasonmp85:
This is pretty problematic. `pg_shard` only checks for constant expressions in clauses involving the partition key, meaning anything goes in other clauses. Take this table:

```sql
CREATE TABLE limit_orders (
  id bigint PRIMARY KEY,
  symbol text NOT NULL,
  bidder_id bigint NOT NULL,
  placed_at timestamp NOT NULL,
  kind order_side NOT NULL,
  limit_price decimal NOT NULL DEFAULT 0.00 CHECK (limit_price >= 0.00)
);

SELECT master_create_distributed_table('limit_orders', 'id');
```

Trying to `INSERT` into this table using non-constant expressions will fail:

```sql
INSERT INTO limit_orders VALUES (2036, 'GOOG', 5634, now(), 'buy', random());
# ERROR:  cannot plan sharded modification containing values which are not constants or constant expressions
```

But `pg_shard` happily accepts a `random()` appearing in a `DELETE` command's clauses:

```sql
DELETE FROM limit_orders WHERE id = 246 AND bidder_id = (random() * 1000);
```

I believe the root cause here is the fact that `UPDATE` and `DELETE` share code with `SELECT`. It's _probably_ OK that `SELECT` accepts `random()` in its `WHERE` clause: for `SELECT` queries, we need only execute on a single replica, so the fact that calls to `random()` on two replicas will return different values is irrelevant.

But for `DELETE` and `UPDATE` this is an issue: replicas will diverge. We should either always require constant expressions or work towards evaluating function calls on the master before executing remote portions of the plan.

PR #51, opened 2015-01-12
Improve unsupported feature identification - Subqueries, CTEs and Functions

### LINKED ISSUE #14: Misleading errors in queries with subselects
_From @samay-sharma on November 21, 2014 14:56_

I am reporting this mainly because of the query it sends to the worker node. I don't think this is a high priority but we should show a clearer error message on the master.

Steps to reproduce:
```sql
SELECT count(*) FROM lineitem WHERE l_orderkey = (SELECT min(l_orderkey) FROM lineitem);
# WARNING:  Bad result from ip-10-102-169-241.ec2.internal:5432
# DETAIL:  Remote message: there is no parameter $0
# WARNING:  Bad result from ip-10-181-241-37.ec2.internal:5432
# DETAIL:  Remote message: there is no parameter $0
# ERROR:  could not receive query results
```

Worker node log:
```sql
# LOG:  statement: SELECT l_orderkey FROM ONLY lineitem_32 WHERE (l_orderkey = $0)
# ERROR:  there is no parameter $0 at character 61
# STATEMENT:  SELECT l_orderkey FROM ONLY lineitem_32 WHERE (l_orderkey = $0)
```

The error message while having such a `WHERE` condition in an `UPDATE` statement is also misleading.
```sql
# Showing min value
SELECT min(l_orderkey) FROM lineitem;
#  min 
# -----
#   1
# (1 row)

# This works as expected
UPDATE lineitem SET l_linenumber = 2 WHERE l_orderkey = 1;
# UPDATE 1

# This statement should touch only one shard, so the error message is a bit misleading.
UPDATE lineitem SET l_linenumber = 2 WHERE l_orderkey = (SELECT min(l_orderkey) FROM lineitem);
# ERROR:  cannot modify multiple shards during a single query
```

### LINKED ISSUE #46: Queries with CTEs produce misleading error
This is a pretty simple bug: queries with a CTE generate an error message about JOINs (_Joins are currently unsupported_) rather than one about CTEs. Check out this [test output][] to see what I mean.

I'd expect the error message to say something about CTEs being unsupported.

[test output]: https://github.com/citusdata/pg_shard/blob/fcf844ce828f85f7c3ad769083b3050cc459393f/expected/modifications.out.tmpl#L90-L94

onderkalaci:
Hey @jasonmp85 ,

* For subqueries we need to make two checks, which we discussed earlier:
    1. One for queryTree->hasSubLinks == true
    2. One for rangeTableEntry->rtekind == RTE_SUBQUERY

* For the common table expressions, I added a new check. In the query struct, there is a queryTree->cteList field. Wherever a CTE exits in a query (SELECT,FROM or WHERE), the field is updated.

* Since we do not support CTEs at all, it is OK to check that variable. Also, it becomes unnecessary to check it in range table entries iteration.

* I think we also need to disccuss functions. For the functions which appears in SELECT or WHERE clause, we can plan the query and get correct results (as far as I observed).

* However, for the functions which appears in FROM clause, we fail to plan and execute the query. Thus, I explictly added a check for rangeTableEntry->rtekind == RTE_FUNCTION. Could I be missing anything on that?

* I added unit tests for each case. Also, tests for common table entries were erroring out wrong, reporting that "Joins are unsupported". I also updated that error report too.

fixes #14
fixes #46

Review tasks:

  - [x] Update messages and details with provided text
  - [x] Change tests to use uppercase keywords
  - [x] Remove extraneous space in tests

### DIFF
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -385,16 +385,20 @@ ErrorIfQueryNotSupported(Query *queryTree)
 	if (queryTree->hasSubLinks == true)
 	{
 		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						errmsg("cannot perform distributed planning on this query"),
-						errdetail("Subqueries are currently unsupported.")));
+						errmsg("cannot perform distributed planning for the given"
+							   " query"),
+						errdetail("Subqueries are not supported in distributed"
+								  " queries.")));
 	}
 
 	/* reject queries which include CommonTableExpr */
 	if (queryTree->cteList != NIL)
 	{
 		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						errmsg("cannot perform distributed planning on this query"),
-						errdetail("Common table entries are currently unsupported.")));
+						errmsg("cannot perform distributed planning for the given"
+							   " query"),
+						errdetail("Common table expressions are not supported in"
+								  " distributed queries.")));
 	}
 
 	/* extract range table entries */
@@ -423,16 +427,18 @@ ErrorIfQueryNotSupported(Query *queryTree)
 			char *rangeTableEntryErrorDetail = NULL;
 			if (rangeTableEntry->rtekind == RTE_SUBQUERY)
 			{
-				rangeTableEntryErrorDetail = "Subqueries are currently unsupported.";
+				rangeTableEntryErrorDetail = "Subqueries are not supported in"
+											 " distributed queries.";
 			}
 			else if (rangeTableEntry->rtekind == RTE_JOIN)
 			{
-				rangeTableEntryErrorDetail = "Joins are currently unsupported.";
+				rangeTableEntryErrorDetail = "Joins are not supported in distributed"
+											 " queries.";
 			}
 			else if (rangeTableEntry->rtekind == RTE_FUNCTION)
 			{
-				rangeTableEntryErrorDetail = "Functions are currently unsupported"
-											 " in FROM clause.";
+				rangeTableEntryErrorDetail = "Functions must not appear in the FROM"
+											 " clause of a distributed query.";
 			}
 			else
 			{
@@ -440,7 +446,8 @@ ErrorIfQueryNotSupported(Query *queryTree)
 			}
 
 			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-							errmsg("cannot perform distributed planning on this query"),
+							errmsg("cannot perform distributed planning for the given"
+								   " query"),
 							errdetail("%s", rangeTableEntryErrorDetail)));
 		}
 	}
@@ -449,8 +456,9 @@ ErrorIfQueryNotSupported(Query *queryTree)
 	if (queryTableCount != 1)
 	{
 		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						errmsg("cannot perform distributed planning on this query"),
-						errdetail("Joins are currently unsupported.")));
+						errmsg("cannot perform distributed planning for the given"
+							   " query"),
+						errdetail("Joins are not supported in distributed queries.")));
 	}
 
 	/* reject queries which involve multi-row inserts */
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -389,6 +389,14 @@ ErrorIfQueryNotSupported(Query *queryTree)
 						errdetail("Subqueries are currently unsupported.")));
 	}
 
+	/* reject queries which include CommonTableExpr */
+	if (queryTree->cteList != NIL)
+	{
+		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						errmsg("cannot perform distributed planning on this query"),
+						errdetail("Common table entries are currently unsupported.")));
+	}
+
 	/* extract range table entries */
 	ExtractRangeTableEntryWalker((Node *) queryTree, &rangeTableList);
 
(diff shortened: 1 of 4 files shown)


pg_shard.c
```diff
@@ -378,6 +378,25 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
 	}
 
+	/*
+	 * Reject subqueries which are in SELECT or WHERE clause.
+	 * Queries which include subqueries in FROM clauses are rejected below.
+	 */
+	if (queryTree->hasSubLinks == true)
+	{
+		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						errmsg("cannot perform distributed planning on this query"),
```
jasonmp85: I know you're just copy-pasting this message, but can we change it to: _cannot perform distributed planning for the given query_? The use of _this query_ in an error message isn't common within PostgreSQL.

pg_shard.c
```diff
@@ -378,6 +378,25 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
+						errdetail("Subqueries are currently unsupported.")));
```
jasonmp85: I'd like to rewrite all of these details as _FEATURE(S) is/are not supported in distributed queries._. So here we'd have: _Subqueries are not supported in distributed queries._.

pg_shard.c
```diff
@@ -378,6 +378,25 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
+	}
+
+	/* reject queries which include CommonTableExpr */
+	if (queryTree->cteList != NIL)
+	{
+		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						errmsg("cannot perform distributed planning on this query"),
+						errdetail("Common table entries are currently unsupported.")));
```
jasonmp85: It's _Common table **expressions**_. So _Common table expressions are not supported in distributed queries._.

pg_shard.c
```diff
@@ -394,9 +413,35 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
+			 * Error out for rangeTableEntries that we do not support.
+			 * We do not explicitly specify "in FROM clause" in the error detail
+			 * for the features that we do not support at all (SUBQUERY, JOIN).
+			 * We do not need to check for RTE_CTE because all common table expressions
+			 * are rejected above with queryTree->cteList check.
+			 */
+			char *rangeTableEntryErrorDetail = NULL;
+			if (rangeTableEntry->rtekind == RTE_SUBQUERY)
+			{
+				rangeTableEntryErrorDetail = "Subqueries are currently unsupported.";
```
jasonmp85: _Subqueries are not supported in distributed queries._

pg_shard.c
```diff
@@ -394,9 +413,35 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
+			}
+			else if (rangeTableEntry->rtekind == RTE_JOIN)
+			{
+				rangeTableEntryErrorDetail = "Joins are currently unsupported.";
```
jasonmp85: _Joins are not supported in distributed queries._

pg_shard.c
```diff
@@ -394,9 +413,35 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
+			}
+			else if (rangeTableEntry->rtekind == RTE_FUNCTION)
+			{
+				rangeTableEntryErrorDetail = "Functions are currently unsupported"
```
jasonmp85: OK, so this is the odd one out, I guess… How about:

_Functions must not appear in the FROM clause of a distributed query._

pg_shard.c
```diff
@@ -394,9 +413,35 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
+											 " in FROM clause.";
+			}
+			else
+			{
+				rangeTableEntryErrorDetail = "Unrecognized range table entry.";
+			}
+
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("cannot perform distributed planning on this query"),
```
jasonmp85: _ cannot perform distributed planning for the given query_

pg_shard.c
```diff
@@ -405,7 +450,7 @@ ErrorIfQueryNotSupported(Query *queryTree)
 	{
 		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 						errmsg("cannot perform distributed planning on this query"),
```
jasonmp85: _cannot perform distributed planning for the given query_

pg_shard.c
```diff
@@ -405,7 +450,7 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
-						errdetail("Joins are currently unsupported")));
+						errdetail("Joins are currently unsupported.")));
```
jasonmp85: _Joins are not supported in distributed queries._

sql/queries.sql
```diff
@@ -110,12 +110,27 @@ CREATE TABLE authors ( name text, id bigint );
 WITH long_names AS ( SELECT id FROM authors WHERE char_length(name) > 15 )
 SELECT title FROM articles;
 
--- joins are unsupported
-SELECT title, authors.name FROM authors, articles WHERE authors.id = articles.author_id;
+--  queries which involve functions in FROM clause are unsupported.
+select * from  articles, position('om' in 'Thomas');
```
jasonmp85: Use uppercase for SQL keywords: `SELECT`, `FROM`, etc. As a general rule, mimic the style of whatever file you're editing.

sql/queries.sql
```diff
@@ -110,12 +110,27 @@ CREATE TABLE authors ( name text, id bigint );
...
+-- subqueries are not supported in SELECT clause
+SELECT  a.title AS  name,(SELECT a2.id FROM authors a2 WHERE a.id = a2.id  LIMIT 1) AS special_price FROM articles a;
+
+-- joins are not supported in WHERE clause
+SELECT title, authors.name FROM authors, articles WHERE authors.id = articles.author_id;
+
+-- joins are not supported in FROM clause
+SELECT * FROM  (articles INNER JOIN authors ON articles.id = authors.id);
+
+
```
jasonmp85: If this space is intentional (and intended to separate the file into single- and multi-shard tests), make it two blank lines. We don't use any more than that in our style.

jasonmp85:
Looking over these changes there's very little you need to do before shipping this. Because you're touching all the error messages I've given feedback about those, but other than that, looks good.

As far as functions go, they can appear within the `FROM` clause, either returning single values, a series, or sets of records (a "function scan"). We don't currently support any of these cases (in fact if we find one I'd be surprised if there weren't also a join node somewhere), so the check you've added is correct. If they appear elsewhere (`SELECT`, `WHERE`) the query can be planned and executed, with some caveats (#47).

So address the three checkboxes I've added up top and :shipit:!

PR #54, opened 2015-01-15
Fix multi-shard SELECT to pull only needed columns

### LINKED ISSUE #33: Multi-shard SELECT queries pull unnecessary columns
`pg_shard` is pulling all data from any column named in the `WHERE` clause across the wire, even if that column is not in the target list. If those columns are large, this makes queries which aren't using the partition key very slow due to data transfer time, and results in huge temporary tables on the master.

```sql
select count(*) from members where profile @> '{ "groups" : [{ "name" : "SFPUG" } ] }';
#  count
# --------
#  183891
# (1 row)
#
# Time: 20229.489 ms

select id from members where profile @> '{ "groups" : [ {"name" : "SFPUG" } ] }';
# Time: 20315.859 ms

# select count(*) from members_10008 where profile @> '{"groups" : [ { "name" : "SFPUG" } ] }';
#  count
# -------
#   5689
# (1 row)
#
# Time: 30.092 ms
```

Even if we did all 32 shards serially, that's still only 950ms. So the 20s response time suggests that `pg_shard` is pulling all of the data from all shards even though we're not looking at the tuples.  In fact:

```sql
select count(*) from members;
#   count
# ---------
#  7879817
# (1 row)
#
# Time: 9181.476 ms
```

But this appears to be because qual pushdown is working strangely:

```sql
select id from members where email like 'j%' limit 1000;
# Time: 8607.005 ms

select id from members where email like 'jst%' limit 1000;
# Time: 2651.042 ms
```

So I did some digging, and it turns out that the problem is that `pg_shard` returns too much data, regardless of what's in the `SELECT` list. When the records are JSON blobs or very wide rows, this is a major issue:

```sql
select sum(1) from members where profile @> '{ "groups" : [ { "name" : "SFPUG" } ] }';
#   sum
# --------
#  183891
# (1 row)
```

Notice that there are no columns in the projection list.  What should get returned to the master is just a column of 183,000 integers. But:

```sql
select pg_size_pretty(pg_total_relation_size('pg_shard_temp_table_7220_1'));
#  pg_size_pretty
# ----------------
#  276 MB
# (1 row)

select * from pg_shard_temp_table_7220_1 limit 1;
# id      |
# email   |
# profile | {
#         |     "address": null,
#         |     "country": "Indonesia",
#         |     "created": "2012-03-11T20:00:32Z",
#         |     "email": "areed228o9y@slashdot.org",
#         |     "first_name": "Amanda",
#         |     (omitted for brevity)
#         |     "quote": null
#         | }

select sum(1) from members where email like 'js%';
# -[ RECORD 1 ]
# sum | 76993

select * from pg_shard_temp_table_7220_2 limit 5;
# -[ RECORD 1 ]-------------------
# id      |
# email   | jsanchez20x1bp@cdc.gov
# profile |
# -[ RECORD 2 ]-------------------
# id      |
# email   | jsanchez20x42u@cdc.gov
# profile |
```
… if any column is used in the `WHERE` clause, `pg_shard` pulls that column across the wire even if it's not in the projection list at all.

jasonmp85:
Reworks multi-shard SELECT so instead of performing quals both remotely and on the local sequential scan, they are only performed remotely. This frees us up to avoid pulling columns mentioned only in quals over the wire, which provides users more control over how much data gets pulled in a multi-shard query.

fixes #33

### DIFF
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -98,6 +98,8 @@ static bool SelectFromMultipleShards(Query *query, List *queryShardList);
 static Query * BuildLocalQuery(Query *query, List *localRestrictList);
 static PlannedStmt * PlanSequentialScan(Query *query, int cursorOptions,
 										ParamListInfo boundParams);
+static void ClassifyRestrictions(List *queryRestrictList, List **remoteRestrictList,
+                                 List **localRestrictList);
 static Query * RowAndColumnFilterQuery(Query *query, List *remoteRestrictList,
                                        List *localRestrictList);
 static List * QueryRestrictList(Query *query);
@@ -261,8 +263,9 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
 			List *remoteRestrictList = NIL;
 			List *localRestrictList = NIL;
 
-			/* TODO: replace with call to classify restrictions */
-			remoteRestrictList = queryRestrictList;
+			/* partition restrictions into remote and local lists */
+			ClassifyRestrictions(queryRestrictList, &remoteRestrictList,
+			                     &localRestrictList);
 
 			/* build local and distributed query */
 			distributedQuery = RowAndColumnFilterQuery(distributedQuery,
@@ -660,7 +663,9 @@ SelectFromMultipleShards(Query *query, List *queryShardList)
 
 
 /*
- * TODO: Add documentation
+ * BuildLocalQuery returns a copy of query with its quals replaced by those
+ * in localRestrictList. Expects queries with a single entry in their FROM
+ * list.
  */
 static Query *
 BuildLocalQuery(Query *query, List *localRestrictList)
@@ -668,11 +673,9 @@ BuildLocalQuery(Query *query, List *localRestrictList)
 	Query *localQuery = copyObject(query);
 	FromExpr *joinTree = localQuery->jointree;
 
-	if (joinTree != NULL)
-	{
-		Assert(list_length(joinTree->fromlist) == 1);
-		joinTree->quals = (Node *) make_ands_explicit((List *) localRestrictList);
-	}
+	Assert(joinTree != NULL);
+	Assert(list_length(joinTree->fromlist) == 1);
+	joinTree->quals = (Node *) make_ands_explicit((List *) localRestrictList);
 
 	return localQuery;
 }
@@ -682,7 +685,8 @@ BuildLocalQuery(Query *query, List *localRestrictList)
  * PlanSequentialScan attempts to plan the given query using only a sequential
  * scan of the underlying table. The function disables index scan types and
  * plans the query. If the plan still contains a non-sequential scan plan node,
- * the function errors out.
+ * the function errors out. Note this function modifies the query parameter, so
+ * make a copy before calling PlanSequentialScan if that is unacceptable.
  */
 static PlannedStmt *
 PlanSequentialScan(Query *query, int cursorOptions, ParamListInfo boundParams)
@@ -723,6 +727,41 @@ PlanSequentialScan(Query *query, int cursorOptions, ParamListInfo boundParams)
 	enable_bitmapscan = bitmapScanEnabledOldValue;
 
 	return sequentialScanPlan;
+}
+
+
+/*
+ * ClassifyRestrictions divides a query's restriction list in two: the subset
+ * of restrictions safe for remote evaluation and the subset of restrictions
+ * that must be evaluated locally. remoteRestrictList and localRestrictList are
+ * output parameters to receive these two subsets.
+ *
+ * Currently places all restrictions in the remote list and leaves the local
+ * one totally empty.
+ */
+static void
+ClassifyRestrictions(List *queryRestrictList, List **remoteRestrictList,
+                     List **localRestrictList)
+{
+	ListCell *restrictCell = NULL;
+
+	*remoteRestrictList = NIL;
+	*localRestrictList = NIL;
+
+	foreach(restrictCell, queryRestrictList)
+	{
+		Node *restriction = (Node *) lfirst(restrictCell);
+		bool restrictionSafeToSend = true;
+
+		if (restrictionSafeToSend)
+		{
+			*remoteRestrictList = lappend(*remoteRestrictList, restriction);
+		}
+		else
+		{
+			*localRestrictList = lappend(*localRestrictList, restriction);
+		}
+	}
 }
 
 
@@ -1028,8 +1067,7 @@ BuildDistributedPlan(Query *query, List *shardIntervalList)
 
 		if (LogDistributedStatements)
 		{
-			ereport(LOG, (errmsg("distributed statement: %s", queryString->data),
-						  errhidestmt(true)));
+			ereport(LOG, (errmsg("distributed statement: %s", queryString->data)));
 		}
 
 		task = (Task *) palloc0(sizeof(Task));
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -95,13 +95,13 @@ static Oid ExtractFirstDistributedTableId(Query *query);
 static bool ExtractRangeTableEntryWalker(Node *node, List **rangeTableList);
 static List * DistributedQueryShardList(Query *query);
 static bool SelectFromMultipleShards(Query *query, List *queryShardList);
-static Query * BuildLocalQuery(Query *query, List *localRestrictList);
-static PlannedStmt * PlanSequentialScan(Query *query, int cursorOptions,
-										ParamListInfo boundParams);
 static void ClassifyRestrictions(List *queryRestrictList, List **remoteRestrictList,
                                  List **localRestrictList);
 static Query * RowAndColumnFilterQuery(Query *query, List *remoteRestrictList,
                                        List *localRestrictList);
+static Query * BuildLocalQuery(Query *query, List *localRestrictList);
+static PlannedStmt * PlanSequentialScan(Query *query, int cursorOptions,
+										ParamListInfo boundParams);
 static List * QueryRestrictList(Query *query);
 static Const * ExtractPartitionValue(Query *query, Var *partitionColumn);
 static bool ExtractFromExpressionWalker(Node *node, List **qualifierList);
@@ -663,74 +663,6 @@ SelectFromMultipleShards(Query *query, List *queryShardList)
 
 
 /*
- * BuildLocalQuery returns a copy of query with its quals replaced by those
- * in localRestrictList. Expects queries with a single entry in their FROM
- * list.
- */
-static Query *
-BuildLocalQuery(Query *query, List *localRestrictList)
-{
-	Query *localQuery = copyObject(query);
-	FromExpr *joinTree = localQuery->jointree;
-
-	Assert(joinTree != NULL);
-	Assert(list_length(joinTree->fromlist) == 1);
-	joinTree->quals = (Node *) make_ands_explicit((List *) localRestrictList);
-
-	return localQuery;
-}
-
-
-/*
- * PlanSequentialScan attempts to plan the given query using only a sequential
- * scan of the underlying table. The function disables index scan types and
- * plans the query. If the plan still contains a non-sequential scan plan node,
- * the function errors out. Note this function modifies the query parameter, so
- * make a copy before calling PlanSequentialScan if that is unacceptable.
- */
-static PlannedStmt *
-PlanSequentialScan(Query *query, int cursorOptions, ParamListInfo boundParams)
-{
-	PlannedStmt *sequentialScanPlan = NULL;
-	bool indexScanEnabledOldValue = false;
-	bool bitmapScanEnabledOldValue = false;
-	List *rangeTableList = NIL;
-	ListCell *rangeTableCell = NULL;
-
-	/* error out if the table is a foreign table */
-	ExtractRangeTableEntryWalker((Node *) query, &rangeTableList);
-
-	foreach(rangeTableCell, rangeTableList)
-	{
-		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) lfirst(rangeTableCell);
-		if (rangeTableEntry->rtekind == RTE_RELATION)
-		{
-			if (rangeTableEntry->relkind == RELKIND_FOREIGN_TABLE)
-			{
-				ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-								errmsg("select from multiple shards is unsupported "
-									   "for foreign tables")));
-			}
-		}
-	}
-
-	/* disable index scan types */
-	indexScanEnabledOldValue = enable_indexscan;
-	bitmapScanEnabledOldValue = enable_bitmapscan;
-
-	enable_indexscan = false;
-	enable_bitmapscan = false;
-
-	sequentialScanPlan = standard_planner(query, cursorOptions, boundParams);
-
-	enable_indexscan = indexScanEnabledOldValue;
-	enable_bitmapscan = bitmapScanEnabledOldValue;
-
-	return sequentialScanPlan;
-}
-
-
-/*
  * ClassifyRestrictions divides a query's restriction list in two: the subset
  * of restrictions safe for remote evaluation and the subset of restrictions
  * that must be evaluated locally. remoteRestrictList and localRestrictList are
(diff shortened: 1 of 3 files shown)


pg_shard.c
```diff
@@ -625,8 +634,13 @@ PlanSequentialScan(Query *query, int cursorOptions, ParamListInfo boundParams)
...
 	enable_bitmapscan = false;
 
-	queryCopy = copyObject(query);
-	sequentialScanPlan = standard_planner(queryCopy, cursorOptions, boundParams);
+	/*
+	 * Remove all quals from query before calling planner. Remote workers will
+	 * apply quals for us: RowAndColumnFilterQuery ensures all quals are pushed
+	 * down during remote executions. The mutator returns a copy of the query.
+	 */
+	modifiedQuery = (Query *) RemoveFromExpressionQualsMutator((Node *) query, NULL);
```
sumedhpathak: Why do we need the mutator construct here? Isn't setting the quals to NIL not sufficient?

pg_shard.c
```diff
@@ -813,6 +818,46 @@ ExtractFromExpressionWalker(Node *node, List **qualifierList)
 
 
 /*
+ * RemoveFromExpressionQualsMutator walks over a query and removes all quals
+ * from any FromExpr nodes it encounters. Returns a copy of the original query.
+ */
+static Node *
+RemoveFromExpressionQualsMutator(Node *originalNode, void *context)
```
sumedhpathak: This PR then doesn't handle the mutable clauses use-case? It appears to remove all the clauses from local evaluation?

jasonmp85: Correct. At the moment I'm removing _all_ local clauses and continue to send all clauses for remote execution.

#47 uses the word `VOLATILE` in its description, but the intent is closer to what you just described. Because the fix for what you describe needs to address the problem for `UPDATE` and `DELETE` statements as well (and possibly `INSERT`?) it will be better handled in a PR for #47 which will be focused entirely on that problem.

pg_shard.c
jasonmp85: The `FromExpr` is actually a tree, so recursing down it to set all `quals` to `NULL` is most appropriate. I believe that we shouldn't ever see anything other than a lone node as we're not supporting `JOINs` at the moment, but it seemed more "correct" to recurse along the entire tree.

I could set `quals` to `NULL` and just document with a `TODO` or something that we would need to change this piece of code if we ever push down `JOIN`s.…

sumedhpathak:
We discussed about modifying the code to make it closer to postgres_fdw, which means we break up quals into 'local' and 'remote'.
The remote query would then pull columns for the local quals, and execute the remote quals entirely. This makes the logic cleaner, and sets it up for handling volatile/mutable functions. To begin with, we expect the local quals to be NIL as we push all of them down currently.

We also discussed using a mutator for modifying the FromExpr for the local quals, vs setting the quals to NULL. We decided on taking the simpler approach now, acknowledging that this may need to be revisited with subselect handling.

jasonmp85:
I haven’t written the `ClassifyRestrictions` function yet, though at the moment it will trivially put all quals in the remote list and leave the local list empty. So all the functionality is there to use these variables (even if the local quals eventually have something to do), but that function is absent (as is documentation).

I didn’t want to change PlanSequentialScan’s parameter list as they’re analogous to the ordering of params in other `plan*` functions in PostgreSQL.

Finally, I want to point out that our old code had weird requirements around what received the original `query` and what received `distributedQuery`, and in fact certain lists and objects build using `distributedQuery` were later used with `query` to build the sequential scan, which felt weird. I retained all those requirements, as they're actually necessary, but it bears calling out that `distributedQuery` has had mutations performed on it by `standard_planner` which means the final `localQuery` is a bit of an amalgam of computed and raw things. Again, this is preserving existing behavior of `pg_shard` but I wanted to mention it.

pg_shard.c
```diff
@@ -937,6 +973,12 @@ BuildDistributedPlan(Query *query, List *shardIntervalList)
 
 		deparse_shard_query(query, shardId, queryString);
 
+		if (LogDistributedStatements)
+		{
+			ereport(LOG, (errmsg("distributed statement: %s", queryString->data),
+						  errhidestmt(true)));
```
sumedhpathak: The comment for errhidestmt mentions that it can be used for error messages which contain the statement. This one doesn't contain the original statement (it contains the transformed query that we send down to the worker nodes). Is it useful to retain that, so we can tie which transformed statement was a result of which original user query?

pg_shard.c
```diff
@@ -641,14 +679,13 @@ PlanSequentialScan(Query *query, int cursorOptions, ParamListInfo boundParams)
...
-RowAndColumnFilterQuery(Query *query)
+RowAndColumnFilterQuery(Query *query, List *remoteRestrictList, List *localRestrictList)
 {
 	Query *filterQuery = NULL;
 	List *rangeTableList = NIL;
-	List *whereClauseList = NIL;
-	List *whereClauseColumnList = NIL;
-	List *projectColumnList = NIL;
 	List *columnList = NIL;
+	List *localColumnList = NIL;
```
sumedhpathak: Can we think of other, more descriptive names here? LocalColumnList doesn't immediately indicate to me what it means. What about retaining whereClauseColumnList?

pg_shard.c
```diff
@@ -703,8 +739,8 @@ RowAndColumnFilterQuery(Query *query)
 	filterQuery = makeNode(Query);
 	filterQuery->commandType = CMD_SELECT;
 	filterQuery->rtable = rangeTableList;
-	filterQuery->targetList = targetList;
 	filterQuery->jointree = fromExpr;
+	filterQuery->targetList = targetList;
```
sumedhpathak: Any reason for this change? (Changing the ordering)

pg_shard.c
```diff
@@ -92,9 +95,11 @@ static Oid ExtractFirstDistributedTableId(Query *query);
 static bool ExtractRangeTableEntryWalker(Node *node, List **rangeTableList);
 static List * DistributedQueryShardList(Query *query);
 static bool SelectFromMultipleShards(Query *query, List *queryShardList);
+static Query * BuildLocalQuery(Query *query, List *localRestrictList);
```
sumedhpathak: I've used a heuristic of trying to maintain the order based on order of the call to the function. So in this case RowAndColumnFilterQuery is being called before BuildLocalQuery. Could you maintain that heuristic (and also make sure the forward declaration order is the same as the implementation order?

pg_shard.c
```diff
@@ -179,6 +184,11 @@ _PG_init(void)
 							 &UseCitusDBSelectLogic, false, PGC_USERSET, 0, NULL,
 							 NULL, NULL);
 
+	DefineCustomBoolVariable("pg_shard.log_distributed_statements",
```
sumedhpathak: Naming: log_remote_query? log_shard_query? What do you think of those options?

pg_shard.c
```diff
@@ -246,15 +256,26 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
 		if (selectFromMultipleShards)
 		{
 			Oid distributedTableId = InvalidOid;
+			Query *localQuery = NULL;
+			List *queryRestrictList = QueryRestrictList(distributedQuery);
+			List *remoteRestrictList = NIL;
+			List *localRestrictList = NIL;
+
+			/* TODO: replace with call to classify restrictions */
```
sumedhpathak: Clarification: Do you intend to make this change in the current PR? Or is this for the future?

pg_shard.c
```diff
@@ -246,15 +256,26 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
...
+			remoteRestrictList = queryRestrictList;
+
+			/* build local and distributed query */
+			distributedQuery = RowAndColumnFilterQuery(distributedQuery,
```
sumedhpathak: Should we rename to remoteQuery as we're using localQuery? local/remote have a more natural pairing than local/distributed.

pg_shard.c
```diff
@@ -586,6 +607,25 @@ SelectFromMultipleShards(Query *query, List *queryShardList)
 
 
 /*
+ * TODO: Add documentation
+ */
+static Query *
+BuildLocalQuery(Query *query, List *localRestrictList)
```
sumedhpathak: Function naming: Do we expect this function to do anything apart from updating the where-clause list? If so, should we use a more specific name here? Something like UpdateRestrictClauses()?

pg_shard.c
```diff
@@ -586,6 +607,25 @@ SelectFromMultipleShards(Query *query, List *queryShardList)
...
+{
+	Query *localQuery = copyObject(query);
+	FromExpr *joinTree = localQuery->jointree;
+
+	if (joinTree != NULL)
```
sumedhpathak: Should this be an assert? If the joinTree is NULL, we aren't updating the clauses, but returning the query as-is. We should either: a) Error out if we expect joinTree to be NULL, b) Assert if we expect it to be there, c) Create a joinTree if its NULL and set the clauses?

pg_shard.c
```diff
@@ -625,8 +664,7 @@ PlanSequentialScan(Query *query, int cursorOptions, ParamListInfo boundParams)
 	enable_indexscan = false;
 	enable_bitmapscan = false;
 
-	queryCopy = copyObject(query);
-	sequentialScanPlan = standard_planner(queryCopy, cursorOptions, boundParams);
+	sequentialScanPlan = standard_planner(query, cursorOptions, boundParams);
```
sumedhpathak: So the reason we kept this here is because the standard_planner scribbles on the query input. This way, I know PlanSequentialScan will keep my inputs unchanged. Removing this means the caller needs to be aware of this, and create a copy of the query in advance before calling this function. I'd be in favor of keeping it here.

jasonmp85: `BuildLocalQuery` creates a copy for us, but I didn't want to put its logic inside of `PlanSequentialScan`. So we're left with either:

  * Having two calls to `copyObject` even though only one is needed
  * Moving `BuildLocalQuery`'s logic inside of `PlanSequentialScan`, which complicates the signature and single responsibility principle.

What if I just change the function comment on `PlanSequentialScan` to document that it expects to "own" `query` and will modify it?

pg_shard.c
jasonmp85: I _think_ this is so disparate callers can ensure a given _STATEMENT:_ line is only emitted once per stack level. I've looked into my use of it here and it serves no purpose. I'm not suppressing the _STATEMENT:_ line, nor am I trying to print more than one line per shard, so it's not suppressing anything ever. I'll remove it.

jasonmp85: BTW the config to emit the "main" query is `log_statement` and is orthogonal to this line.

pg_shard.c
jasonmp85: They were named after `localConds` and `remoteConds` in `postgres_fdw`. I find the `local`/`remote` distinction _very_ explicit/clear, as it immediately tells me what a variable is used for (local or remote execution). `whereClause` says where we got the value, not what it is for.

But while writing this I remember wondering if the naming couldn't be better, so I want to dump all the `local`/`remote` names here and we can discuss which need changing:

  * `localQuery` — The `Query` structure used to build a local sequential scan
  * `localRestrictList` — A list of expressions from `WHERE` clauses that must be included in the local sequential scan. Currently always `NIL`.
  * `localColumnList` — The full list of columns referenced by the `localRestrictList` quals
  * `remoteRestrictList` — A list of expressions from `WHERE` clauses that will be executed remotely. Currently contains all `WHERE` clauses from the original query.

Was it just `localColumnList` you had a problem with, or are any of the others problematic? Some of these occur both as variable names and as parameter names.

pg_shard.c
jasonmp85: Yes. The above blocks build `rangeTableList` then `fromExpr` then `targetList`. I felt the assignments should mirror the order of construction.

pg_shard.c
jasonmp85: I indeed wanted to change the order but knew doing so would dirty the diff (it would just highlight the whole method in green). You could always view each commit separately, but that makes commenting harder. I'm going to clean them up in exactly this way once we hash out the implementation issues first.

pg_shard.c
jasonmp85: The logic I used to get me `log_distributed_statements`

  * Some settings begin with `debug_` and others with `log_`, though the debug ones tend to be about logging a complicated structure, not one line. So I went with `log_`
  * The (probably well-known) configuration for PostgreSQL to log every query is `log_statement`. I therefore wanted to use _statement_ in my name
  * As an end-user I understand _query_ to mean a `SELECT`: something that asks for data and returns it. A _statement_, however, has side-effects
  * `distributed` is because these statements are fields in the distributed plan. I'm logging them at the point of planning, not execution, so they are not yet remote (`log_statement` also logs far before execution)
  * Finally, it's plural because unlike `log_statement`, we're going to have multiple statements per plan

So in short, I don't like _query_ (inappropriate for `INSERT`, `UPDATE`, or `DELETE`) and `remote` makes sense if I'm logging in the course of remote execution, which I'm not. Tell me if this isn't convincing.

pg_shard.c
jasonmp85: Right now.

pg_shard.c
jasonmp85: Hm. I honestly don't know. The problem is the object is a `DistributedPlan`, so there's a duality in that we build the `distributedPlan` using the `distributedQuery` but use the `localQuery` to build a local plan (the sequential scan, currently just named `plannedStatement`). I'll have to think a bit more about this.

pg_shard.c
jasonmp85: It doesn't modify the input, it returns a modified copy (this ties into the question about `copyObject` later).

pg_shard.c
jasonmp85: You're right: I was focusing on the `fromlist` having exactly one entry, but we can also expect `joinTree` to never be null. Going to change to assert (don't want to error as this should not be a user-facing problem: `Assert` is more appropriate).

jasonmp85:
Responded to these pieces of feedback:
  * Remove use of `errhidestmt`
  * Added Assert that `query->jointree` is never `NULL`
  * Documented `BuildLocalQuery`
  * Clarified that `PlanSequentialScan` modifies its `query` argument
  * Added `ClassifyRestrictions` function

Needing further feedback:

  - [ ] Clarification on `local`/`remote` naming, as it's somewhat pervasive
  - [ ] How against `log_distributed_statements` are you? I like it and find alternatives less clear
  - [ ] How should I address the `copyObject` question? I added comments clarifying that one function _is_ returning a copy and another will scribble on input, but I don't see any other solution without adding an extraneous copy or adding too much responsibility to `PlanSequentialScan`. What was your suggestion?
  - [ ] Look over `ClassifyRestrictions`. It's naïve for now but I'll fill it in as part of #47
  - [ ] Is _restriction_ the right word? I like _qual_ but it feels informal (and _qualifier_ or _qualification_ feel too formal). What do we call the expressions in a `WHERE` clause? Couldn't find clear precedent in Citus…

jasonmp85:
(And as mentioned, I'll modify function declaration/definition order as a last step).

pg_shard.c
sumedhpathak: Hm, I see. I was under the impression this was only being logged for SELECT's, but looks like its logging even updates and deletes. I'm convinced then to leave it as it is.

pg_shard.c
sumedhpathak: Hmm, so now since we modify the localQuery, it makes sense to make a copy of it before we update the quals to NIL. Lets change the function comment here to document that it modifies the inputs. On looking at it a bit more, I like that the planner modifications to the localQuery which generated the plannedStmt are available to the caller of this function.

pg_shard.c
sumedhpathak: I think `localQuery`, `localRestrictList` and `remoteRestrictList` all make sense. I think the attribute of `local` to a column isn't immediately descriptive to me. I guess the long form would be localRestrictClauseColumnList? 

In the context of this function, I was thinking the two sources of columns to be fetched: projectColumns and whereClauseColumns. This duality to me makes it a bit clearer. The combination of these two lists from the toBeFetchedColumnList.

What do you think?

sumedhpathak:
I think `restriction` is OK for now. We already have a function called QueryRestrictList, and it has precedence in Postgres. I think we've used whereClause or selectClause in Citus previously.

pg_shard.c
jasonmp85: `toBeFetchedColumnList` is a bit unwieldy, but the others seem great. Going to do that.

pg_shard.c
jasonmp85: Chatted and keeping as `distributedQuery`.

pg_shard.c
jasonmp85: Actually updated the comment already before you wrote this reply, so I think this is covered now.

pg_shard.c
jasonmp85: Calling it `requiredColumnList`, as they're required to service the query.

PR #55, opened 2015-01-16
Fix multi-shard SELECT problem with HAVING clauses

### LINKED ISSUE #53: Queries with HAVING can return incorrect results
Assume a simple table distributed along an `id` column:

```sql
CREATE TABLE employees (
    id integer,
    name text
);

SELECT master_create_distributed_table('employees', 'id');
SELECT master_create_worker_shards('employees', 2, 1);
```

Using `HAVING` clauses in `SELECT` queries works correctly so long as columns referenced in the `HAVING` clause are in a `WHERE`, `ORDER BY`, or `GROUP BY` clause:

```sql
SELECT substr(name, 1, 1) AS init, 
       sum(id)            AS id_total 
FROM   employees 
GROUP  BY init 
HAVING SUM(id) > 200000;
# ┌──────┬──────────┐
# │ init │ id_total │
# ├──────┼──────────┤
# │ H    │   319464 │
# │ M    │  1253983 │
# │ J    │   844768 │
# │ V    │   280801 │
# │ R    │   610906 │
# │ L    │  1166332 │
# │ D    │   829092 │
# │ T    │   700698 │
# │ I    │   201319 │
# │ F    │   284724 │
# │ N    │   363474 │
# │ G    │   401067 │
# │ E    │   697631 │
# │ B    │   526513 │
# │ C    │  1148864 │
# │ S    │  1010327 │
# │ P    │   244061 │
# │ K    │   748454 │
# │ A    │  1079310 │
# └──────┴──────────┘
# (19 rows)
```

But if columns in the `HAVING` clause do _not_ appear elsewhere, the query returns incorrect results:

```sql
SELECT substr(name, 1, 1) AS init 
FROM   employees 
GROUP  BY init 
HAVING SUM(id) > 200000; 
# ┌──────┐
# │ init │
# ├──────┤
# └──────┘
# (0 rows)
```

This is probably related to #33, since it appears `pg_shard` currently only considers quals and projections in deciding which columns to retrieve from the remote. The `HAVING` clause is evaluated locally, so if its columns are not retrieved it cannot be evaluated correctly.

We should either reject the use of `HAVING` or make sure it has the columns it needs.

jasonmp85:
Because `Var`s from `HAVING` clauses weren't being included in the remote query's target list, they would be `NULL` if they did not appear in the target list or a `GROUP`/`ORDER` clause. This change includes them in the remote target list so they'll be available for local evaluation of the `HAVING` clause.

fixes #53

### DIFF
--- a/expected/queries.out
+++ b/expected/queries.out
@@ -193,6 +193,7 @@ SELECT COUNT(*) FROM articles;
     50
 (1 row)
 
+-- try query with more SQL features
 SELECT author_id, sum(word_count) AS corpus_size FROM articles
 	GROUP BY author_id
 	HAVING sum(word_count) > 25000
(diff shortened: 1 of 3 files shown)


sumedhpathak:
Looks good. :shipit:

Issue #49, opened 2014-12-27
Prevent deletion of metadata when still in use


jasonmp85:
Running `DROP EXTENSION` will remove `pg_shard`'s metadata tables, including information about the placement and health of shards. This behavior should only be possible with the `CASCADE` modifier: otherwise the extension should refuse to be dropped until all distributed tables have been removed.

Perhaps we can look into the dependency tracking available in PostgreSQL's catalog tables to satisfy this need.

jasonmp85:
@onderkalaci, in case you pick this up, you might want to look at [`recordDependencyOnSingleRelExpr`][1] and [`recordDependencyOn`][2], which are helpers for making changes to the [`pg_depend`][3] catalog.

[1]: https://github.com/postgres/postgres/blob/REL9_4_0/src/backend/catalog/dependency.c#L1343-L1430
[2]: https://github.com/postgres/postgres/blob/REL9_4_0/src/backend/catalog/pg_depend.c#L36-L49
[3]: http://www.postgresql.org/docs/9.4/static/catalog-pg-depend.html

PR #56, opened 2015-01-20
Add function to sync metadata to CitusDB

### LINKED ISSUE #26: Improve metadata synchronization
Give Citus users a general UDF to sync pg_shard metadata with Citus catalogs.

jasonmp85:
I went back and forth as to whether this should be included with `pg_shard` proper or implemented in a separate extension but ultimately decided the former would be too complicated. Besides, we already have a GUC for CitusDB-related behavior so this isn't setting a new precedent re: encapsulation or anything.

I investigated whether making the C function entirely independently of the rest of the code was worthwhile, but as this is all in one extension that seemed a bit extreme. Another avenue I pursued was moving the longer PL/pgSQL function out to its own file. [This blog post][1] gave me some ideas, but I think they're not worth implementing at the moment.

So I cleaned up whitespace, added function comments, added `UPSERT`-like logic, and ensured we handled `NULL` inputs at all points in the call chain. I'm writing the rest of my unit tests now and will push them up shortly.

fixes #26.

[1]: http://blog.pgxn.org/post/4783001135/extension-makefiles


citus_metadata_sync.c
```diff
@@ -0,0 +1,43 @@
...
+
+#include "citus_metadata_sync.h"
+#include "distribution_metadata.h"
+
+#include "utils/builtins.h"
+
+
+/* declarations for dynamic loading */
+PG_FUNCTION_INFO_V1(partition_column_to_node_string);
+Datum
```
sumedhpathak: Could you add spacing between the declaration, and also more importantly add a function comment?

pg_shard--1.0.sql
```diff
@@ -77,3 +77,78 @@ CREATE FUNCTION worker_copy_shard_placement(table_name text, source_node_name te
...
+		-- copy shard metadata
+		INSERT INTO pg_dist_shard
+					(shardid,
+					 logicalrelid,
+					 shardstorage,
+					 shardminvalue,
+					 shardmaxvalue)
+		SELECT id,
+			   relation_id,
+			   STORAGE,
```
sumedhpathak: Where is this STORAGE parameter coming from? I don't see it defined as a plpgsql variable?

pg_shard--1.0.sql
```diff
@@ -77,3 +77,78 @@ CREATE FUNCTION worker_copy_shard_placement(table_name text, source_node_name te
...
+COMMENT ON FUNCTION partition_column_to_node_string(oid)
+		IS 'return textual form of distributed table''s partition column';
+
+CREATE FUNCTION sync_table_metadata_to_citus(table_name text) RETURNS VOID
+AS $sync_table_metadata_to_citus$
+	DECLARE
+		table_relation_id CONSTANT oid NOT NULL := table_name::regclass::oid;
+		dummy_shard_length CONSTANT bigint := 0;
+	BEGIN
+		-- remove shard placements for table
```
sumedhpathak: I had a few higher-level questions on the upsert functionality:

- We expect this to change if:
1. Someone drops the table, re-creates it, and calls master_create_distributed_table(), and master_create_worker_shards() again. This is what Neustar was doing.

2. Someone only removes the metadata in pgs_distribution_metadata for the table, and re-creates shards using master_create_worker_shards(). 

In case 1, the relationId would have changed, thus the delete should be a no-op. There wouldn't be any existing shards in citusDB's metadata for that new relationId.

In case 2, the partition information might be there, but the shard/shard-placement deletion will be no-ops as the metadata won't exist in pgs_distribution_metadata.

If my understanding is correct, could we remove the delete functionality for now and think about this in a later change? (Maybe as part of the more comprehensive two-way sync item?)

My other question was: If we do a metadata-sync and then call drop-table, do you know if CitusDB complains about adding CASCADE, and if it deletes the entry in pg_dist_partition? (It has this behavior for CitusDB distributed tables, which it should start assuming a pg_shard table is after the metadata sync).

pg_shard--1.0.sql
jasonmp85: That's the stupid SQL formatter deciding `STORAGE` is a keyword (which it is in certain contexts) and capitalizing it. It's the name of a column in the `pg_shard` tables. I'll make it lowercase to be clearer.

pg_shard--1.0.sql
jasonmp85: I'll remove the `DELETE` functionality. It was useful for me during testing, but I see it's either no necessary or incomplete for the end user.

I'll see about `CASCADE`. I didn't know Citus did that. What is the action here? Just to find out whether it happens?

citus_metadata_sync.c
jasonmp85: Done.

sumedhpathak:
:shipit:

pg_shard--1.0.sql
jasonmp85: Rebased this branch and removed the `DELETE`-related commit, so it's like it was never there.

As far as `CASCADE` goes… it doesn't require it. This is because CitusDB calls `recordDependencyOnSingleRelExpr` and `recordDependencyOn` inside of its internal function that creates the partition entry. Once #49 is resolved `pg_shard` itself will require `CASCADE` to drop a table that's still known as distributed. Additionally, CitusDB could change its implementation to use an `INSERT`/`UPDATE` trigger so even a simple `INSERT` into its metadata tables will have dependency information recorded correctly…

PR #57, opened 2015-01-20
pg_shard performance improvement

### LINKED ISSUE #50: Optimize operator expression construction
[`MakeOpExpression`][1] is used to build:

  * Equality comparisons against a hash value which replace other equality constraints in `SELECT`, `UPDATE`, and `DELETE` queries
  * Synthetic clauses used as the hash value of the partition key of an `INSERT`ed row
  * "Between" comparisons which represent the range of hash values contained within a given shard

@onderkalaci has extensively benchmarked performance characteristics of `pg_shard` and determined that avoiding [this call][] to `GetDefaultOpClass` results in a huge performance boost, especially on multi-socket machines (probably due to lock contention).

We previously replaced all calls to `GetDefaultOpClass` with equivalent calls to `lookup_type_cache`, but the type cache does not contain the `≥` and `≤` operators needed for the "between" comparison in the above list. At the time we had already returned substantial improvements in other places, so we left this final call site, but now it is apparent we should remove it as well.

I had previously tried building an equivalent `OR` clause using the greater (or less) than operator and an equality clause (thus enabling the use of simple typecache lookups), but that has performance implications if the number of shards is large.

So the straightforward option here is to just cache the `≥` and `≤` operators ourselves (or cache the entire expression node returned by [`BuildBaseConstraint`][2] and clone it when needed). The need for equality operators can be directly fulfilled by `lookup_type_cache` responses and we can get rid of all remaining calls to `GetDefaultOpClass` (and stop doing duplicate work looking up operators for `int32` during every query).

[1]: https://github.com/citusdata/pg_shard/blob/46bcf2c0b1a5250a286e5dd6b0e9fa42e275725f/prune_shard_list.c#L203-L238
[this call]: https://github.com/citusdata/pg_shard/blob/46bcf2c0b1a5250a286e5dd6b0e9fa42e275725f/prune_shard_list.c#L249
[2]: https://github.com/citusdata/pg_shard/blob/46bcf2c0b1a5250a286e5dd6b0e9fa42e275725f/prune_shard_list.c#L147-L169

onderkalaci:
This branch implements a cache in front of  *GetOperatorByType* function, so that it prevents unnecessary calls to *GetDefaultOpClass* function.

With this implementation, we get the same performance improvement that we observed with @ozgune. I may send the details of the results via an email.

fixes #50

Review tasks:
  - [x] Replace appearances of `OperatorType` with `OperatorId` (also check comments)
  - [x] Short-circuit without caching if `GetOperatorByType` returns `InvalidOid`
  - [x] Clean up the `if` statement with the three boolean expressions
  - [x] Ensure function declaration order matches definition order

### DIFF
--- a/prune_shard_list.c
+++ b/prune_shard_list.c
@@ -43,16 +43,16 @@
 
 
 /*
- * OperatorTypeCache is used for caching base operator types for given typeId,
+ * OperatorIdCache is used for caching operator identifiers for given typeId,
  * accessMethodId and strategyNumber. It is initialized to empty list as
  * there are no items in the cache.
  */
-static List *OperatorTypeCache = NIL;
+static List *OperatorIdCache = NIL;
 
 
 /* local function forward declarations */
-static Oid GetOperatorByType(Oid typeId, Oid accessMethodId, int16 strategyNumber);
 static Oid LookupOperatorByType(Oid typeId, Oid accessMethodId, int16 strategyNumber);
+static Oid GetOperatorByType(Oid typeId, Oid accessMethodId, int16 strategyNumber);
 static bool SimpleOpExpression(Expr *clause);
 static Node * HashableClauseMutator(Node *originalNode, Var *partitionColumn);
 static bool OpExpressionContainsColumn(OpExpr *operatorExpression, Var *partitionColumn);
@@ -250,21 +250,22 @@ MakeOpExpression(Var *variable, int16 strategyNumber)
 
 /*
  * LookupOperatorByType is a wrapper around GetOperatorByType that uses a cache
- * to avoid multiple lookups of operators by their types.
+ * to avoid multiple lookups of operators within a single session by their types.
  */
 static Oid
 LookupOperatorByType(Oid typeId, Oid accessMethodId, int16 strategyNumber)
 {
-	OperatorTypeCacheEntry *matchingCacheEntry = NULL;
+	OperatorIdCacheEntry *matchingCacheEntry = NULL;
 	ListCell *cacheEntryCell = NULL;
 
 	/* search the cache */
-	foreach(cacheEntryCell, OperatorTypeCache)
+	foreach(cacheEntryCell, OperatorIdCache)
 	{
-		OperatorTypeCacheEntry *cacheEntry = lfirst(cacheEntryCell);
+		OperatorIdCacheEntry *cacheEntry = lfirst(cacheEntryCell);
 
-		if (cacheEntry->typeId == typeId && cacheEntry->accessMethodId == accessMethodId
-			&& cacheEntry->strategyNumber == strategyNumber)
+		if ((cacheEntry->typeId == typeId) &&
+			(cacheEntry->accessMethodId == accessMethodId) &&
+			(cacheEntry->strategyNumber == strategyNumber))
 		{
 			matchingCacheEntry = cacheEntry;
 			break;
@@ -274,17 +275,25 @@ LookupOperatorByType(Oid typeId, Oid accessMethodId, int16 strategyNumber)
 	/* if not found in the cache, call GetOperatorByType and put the result in cache */
 	if (matchingCacheEntry == NULL)
 	{
-		Oid operatorId = GetOperatorByType(typeId, accessMethodId, strategyNumber);
-		MemoryContext oldContext = MemoryContextSwitchTo(CacheMemoryContext);
+		Oid operatorId;
+		MemoryContext oldContext;
 
-		matchingCacheEntry = palloc0(sizeof(OperatorTypeCacheEntry));
+		operatorId = GetOperatorByType(typeId, accessMethodId, strategyNumber);
+		if (operatorId == InvalidOid)
+		{
+			/* if operatorId is invalid, return and do not cache its value */
+			return operatorId;
+		}
+
+		oldContext = MemoryContextSwitchTo(CacheMemoryContext);
+
+		matchingCacheEntry = palloc0(sizeof(OperatorIdCacheEntry));
 		matchingCacheEntry->typeId = typeId;
 		matchingCacheEntry->accessMethodId = accessMethodId;
 		matchingCacheEntry->strategyNumber = strategyNumber;
-
 		matchingCacheEntry->operatorId = operatorId;
 
-		OperatorTypeCache = lappend(OperatorTypeCache, matchingCacheEntry);
+		OperatorIdCache = lappend(OperatorIdCache, matchingCacheEntry);
 
 		MemoryContextSwitchTo(oldContext);
 	}
--- a/prune_shard_list.c
+++ b/prune_shard_list.c
@@ -275,10 +275,9 @@ LookupOperatorByType(Oid typeId, Oid accessMethodId, int16 strategyNumber)
 	/* if not found in the cache, call GetOperatorByType and put the result in cache */
 	if (matchingCacheEntry == NULL)
 	{
-		Oid operatorId;
-		MemoryContext oldContext;
+		MemoryContext oldContext = NULL;
+		Oid operatorId = GetOperatorByType(typeId, accessMethodId, strategyNumber);
 
-		operatorId = GetOperatorByType(typeId, accessMethodId, strategyNumber);
 		if (operatorId == InvalidOid)
 		{
 			/* if operatorId is invalid, return and do not cache its value */
--- a/prune_shard_list.c
+++ b/prune_shard_list.c
@@ -39,10 +39,20 @@
 #include "utils/errcodes.h"
 #include "utils/lsyscache.h"
 #include "utils/typcache.h"
+#include "utils/memutils.h"
+
+
+/*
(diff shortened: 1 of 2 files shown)


prune_shard_list.c
```diff
@@ -39,10 +39,20 @@
...
+ * OperatorTypeCache is used for caching base operator types for given typeId,
+ * accessMethodId and strategyNumber. It is initialized to empty list as
+ * there are no items in the cache.
+ */
+static List *OperatorTypeCache = NIL;
 
 
 /* local function forward declarations */
 static Oid GetOperatorByType(Oid typeId, Oid accessMethodId, int16 strategyNumber);
+static Oid LookupOperatorByType(Oid typeId, Oid accessMethodId, int16 strategyNumber);
```
jasonmp85: Move this up one line. The declaration and definition orders should match, but these declarations are in another order than the definitions.

prune_shard_list.c
```diff
@@ -239,6 +249,51 @@ MakeOpExpression(Var *variable, int16 strategyNumber)
 
 
 /*
+ * LookupOperatorByType is a wrapper around GetOperatorByType that uses a cache
+ * to avoid multiple lookups of operators by their types.
```
jasonmp85: I'd say _multiple lookups of operators within a single session_, to make it clear that the cache is process-bound and not in shared memory.

prune_shard_list.c
```diff
@@ -239,6 +249,51 @@ MakeOpExpression(Var *variable, int16 strategyNumber)
...
+{
+	OperatorTypeCacheEntry *matchingCacheEntry = NULL;
+	ListCell *cacheEntryCell = NULL;
+
+	/* search the cache */
+	foreach(cacheEntryCell, OperatorTypeCache)
+	{
+		OperatorTypeCacheEntry *cacheEntry = lfirst(cacheEntryCell);
+
+		if (cacheEntry->typeId == typeId && cacheEntry->accessMethodId == accessMethodId
```
jasonmp85: Style-wise, I think we need explicit grouping here because there are more than two expressions. Can you change to:

```c
if ((cacheEntry->typeId == typeId) &&
	(cacheEntry->accessMethodId == accessMethodId) &&
	(cacheEntry->strategyNumber == strategyNumber))
{
```
It'll be more readable with the parallel `cacheEntry` variables lined up like that.

prune_shard_list.c
```diff
@@ -239,6 +249,51 @@ MakeOpExpression(Var *variable, int16 strategyNumber)
...
+	if (matchingCacheEntry == NULL)
+	{
+		Oid operatorId = GetOperatorByType(typeId, accessMethodId, strategyNumber);
+		MemoryContext oldContext = MemoryContextSwitchTo(CacheMemoryContext);
+
+		matchingCacheEntry = palloc0(sizeof(OperatorTypeCacheEntry));
+		matchingCacheEntry->typeId = typeId;
+		matchingCacheEntry->accessMethodId = accessMethodId;
+		matchingCacheEntry->strategyNumber = strategyNumber;
+
```
jasonmp85: Get rid of this blank line.

prune_shard_list.c
```diff
@@ -239,6 +249,51 @@ MakeOpExpression(Var *variable, int16 strategyNumber)
...
+			matchingCacheEntry = cacheEntry;
+			break;
+		}
+	}
+
+	/* if not found in the cache, call GetOperatorByType and put the result in cache */
+	if (matchingCacheEntry == NULL)
+	{
+		Oid operatorId = GetOperatorByType(typeId, accessMethodId, strategyNumber);
+		MemoryContext oldContext = MemoryContextSwitchTo(CacheMemoryContext);
```
jasonmp85: I'm not sure when the inputs might cause this, but `GetOperatorByType` will return `InvalidOid` if the type isn't found. In that case, we don't want to cache anything, so you should check for that case and avoid caching a value. While it's _unlikely_ that we could get `InvalidOid` during one query and then find a valid operator the next, it's probably not 100% impossible, so it'd be good to do.

See [analogous code][1] in distribution_metadata.c for a pattern to follow.

[1]: https://github.com/citusdata/pg_shard/blob/9fa64edd98858ef1a3eaaca460deab7157dc122e/distribution_metadata.c#L90-L115

prune_shard_list.c
```diff
@@ -39,10 +39,20 @@
 #include "utils/errcodes.h"
 #include "utils/lsyscache.h"
 #include "utils/typcache.h"
+#include "utils/memutils.h"
+
+
+/*
+ * OperatorTypeCache is used for caching base operator types for given typeId,
```
jasonmp85: `OperatorType` doesn't make sense as a phrase. It could make sense to say we're caching the operator, or we could be more direct and say we're caching the operator identifier. `OperatorIdCache` makes the most sense.

Update the comment too: not sure what _base_ means in this context. Maybe just say _is used for caching operator identifiers for a given_.

prune_shard_list.h
```diff
@@ -31,6 +31,21 @@
 #define RESERVED_HASHED_COLUMN_ID MaxAttrNumber
 
 
+/*
+ * OperatorTypeCacheEntry contains the information for a cache entry in
```
jasonmp85: Change to `OperatorIdCacheEntry`.

prune_shard_list.h
```diff
@@ -31,6 +31,21 @@
...
+ * operator type cache.
+ */
+typedef struct OperatorTypeCacheEntry
```
jasonmp85: `OperatorIdCacheEntry`.

prune_shard_list.h
```diff
@@ -31,6 +31,21 @@
...
+{
+	/* cache key consists of typeId, accessMethodId and strategyNumber */
+	Oid typeId;
+	Oid accessMethodId;
+	int16 strategyNumber;
+
```
jasonmp85: Again, I don't think this empty line is meaningful, so just remove it.

jasonmp85:
Overall looks good. I think this is a :shipit: after a few changes that I'll add up top.

jasonmp85:
If you're wondering "why the heck does he add the checklists to the top-level comment rather than down below?" it's because adding them up top gives a little progress bar on the Pull Requests page:

screen shot 2015-01-21 at 11 21 19 pm

onderkalaci:
@jasonmp85 I didn't merged this branch and let you do so. I applied check boxes, only not sure about "Short-circuit without caching if GetOperatorByType returns InvalidOid" item.

You suggested me to see  [analogous code][1] in distribution_metadata.c for a pattern. However, we cannot apply it directly as far as I see. Could you check [my implementation][2], is it OK for you?

[1]: https://github.com/citusdata/pg_shard/blob/9fa64edd98858ef1a3eaaca460deab7157dc122e/distribution_metadata.c#L90-L115
[2]:
https://github.com/citusdata/pg_shard/pull/57/files#diff-777ad11485b48996843ea21a64147014R281

jasonmp85:
I said _analogous_, not identical! You did exactly what I was looking for.

I'm going to close this out myself but need to make one change before I do so: our code style is to _always_ initialize variables. The one place it might be OK to do it on another line is when declaring a local `struct` variable, but even then we make sure to [`memset` them with zeros][1] _as soon as possible_ (i.e. after all other declarations are out of the way). So when you declare `operatorId` it should be set to `InvalidOid` and `oldContext` should initially be set to `NULL`.

[1]: https://github.com/citusdata/pg_shard/blob/9fa64edd98858ef1a3eaaca460deab7157dc122e/distribution_metadata.c#L792-L793

jasonmp85:
I said _analogous_, not identical! You did exactly what I was looking for.

I'm going to close this out myself but need to make one change before I do so: our code style is to _always_ initialize variables. The one place it might be OK to do it on another line is when declaring a local `struct` variable, but even then we make sure to [`memset` them with zeros][1] _as soon as possible_ (i.e. after all other declarations are out of the way). So when you declare `operatorId` it should be set to `InvalidOid` and `oldContext` should initially be set to `NULL`.

[1]: https://github.com/citusdata/pg_shard/blob/9fa64edd98858ef1a3eaaca460deab7157dc122e/distribution_metadata.c#L792-L793

prune_shard_list.c
```diff
@@ -239,6 +249,60 @@ MakeOpExpression(Var *variable, int16 strategyNumber)
...
+		{
+			matchingCacheEntry = cacheEntry;
+			break;
+		}
+	}
+
+	/* if not found in the cache, call GetOperatorByType and put the result in cache */
+	if (matchingCacheEntry == NULL)
+	{
+		Oid operatorId;
```
jasonmp85: Always initialize variables! Value types should be set to their "unset" value (so `InvalidOid` for `Oid`) and pointers to `NULL`.

jasonmp85: Here I think it's actually best to reverse the order of the variables and set `operatorId` immediately to the return of `GetOperatorByType`.

jasonmp85:
Looks good. Pushing up the small change I mentioned and merging.

PR #58, opened 2015-01-21
Handle zero-shard SELECT queries

jasonmp85:
I added an `Assert` in a9564eb based on the assumption that if all shards are pruned, something is wrong. I neglected to think about any `SELECT` query with unsatisfiable restrictions. Such queries cause the `Assert` to fail.

That said, even without the `Assert`, the behavior is undesirable: we print _cannot execute select over multiple shards_, which is terribly misleading (especially since we _can_ do that).

This change updates names to be clearer that the multi-shard `SELECT` path is really just a possibility in execution and is actually valid for zero shards. I've added a unit test to verify that such zero-shard queries return the correct result (no rows).

Review tasks:

  - [x] See how single-shard logic handles this if I remove the `ereport` call
  - [x] Investigate whether PostgreSQL has short-circuit logic for unsatisfiable queries
  - [x] Look into adding entirely new case for zero-shard queries, see how much extra code it adds

### DIFF
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -610,9 +610,10 @@ ExtractRangeTableEntryWalker(Node *node, List **rangeTableList)
 
 /*
  * DistributedQueryShardList prunes the shards for the table in the query based
- * on the query's restriction qualifiers, and returns this list. If the function
- * cannot find any shards for the distributed table, it errors out. In other sense,
- * the function errors out or returns a non-empty list.
+ * on the query's restriction qualifiers, and returns this list. It is possible
+ * that all shards will be pruned if a query's restrictions are unsatisfiable.
+ * In that case, this function can return an empty list; however, if the table
+ * being queried has no shards created whatsoever, this function errors out.
  */
 static List *
 DistributedQueryShardList(Query *query)
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -640,9 +640,6 @@ DistributedQueryShardList(Query *query)
 	prunedShardList = PruneShardList(distributedTableId, restrictClauseList,
 									 shardIntervalList);
 
-	/* shouldn't be an empty list, but assert in case something's very wrong */
-	Assert(prunedShardList != NIL);
-
 	return prunedShardList;
 }
 
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -1111,6 +1111,17 @@ PgShardExecutorStart(QueryDesc *queryDesc, int eflags)
 			bool topLevel = true;
 			LOCKMODE lockMode = NoLock;
 			EState *executorState = NULL;
+			bool zeroShardQuery = list_length(distributedPlan->taskList) == 0;
+
+			/* if query involves zero shards, just let it hit local table */
+			if (zeroShardQuery)
+			{
+				Plan *originalPlan = distributedPlan->originalPlan;
+				plannedStatement->planTree = originalPlan;
+
+				PgShardExecutorStart(queryDesc, eflags);
+				return;
+			}
 
 			/* disallow transactions and triggers during distributed commands */
 			PreventTransactionChain(topLevel, "distributed commands");
@@ -1776,10 +1787,7 @@ ExecuteSingleShardSelect(DistributedPlan *distributedPlan, EState *executorState
 	TupleTableSlot *tupleTableSlot = NULL;
 
 	List *taskList = distributedPlan->taskList;
-	if (list_length(taskList) != 1)
-	{
-		ereport(ERROR, (errmsg("cannot execute select over multiple shards")));
-	}
+	Assert(list_length(taskList) == 1);
 
 	task = (Task *) linitial(taskList);
 	tupleStore = tuplestore_begin_heap(false, false, work_mem);
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -113,6 +113,7 @@ static DistributedPlan * BuildDistributedPlan(Query *query, List *shardIntervalL
 /* executor functions forward declarations */
 static void PgShardExecutorStart(QueryDesc *queryDesc, int eflags);
 static bool IsPgShardPlan(PlannedStmt *plannedStmt);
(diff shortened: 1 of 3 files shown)


sumedhpathak:
So the previous behavior if we pruned away all shards was that we'd get an obscure error saying we can't support multi-shard selects?

At a high-level, I feel overloading the multi-shard select logic is a bit confusing. We seem to be doing it because it is convenient (i.e., it seems to handle it automatically), but the rest of the multi-shard logic is unaware that the task list could be zero.

I had two questions:
1. Does the single-shard logic not handle it if all the tasks are empty? (assuming we removed/changed the existing assert).
2. The higher-level question being, should we have a different code-path for zeroShardSelectLogic? (naming TBD). That way if there are no tasks to execute, we don't have to go through the process of building a local plan, creating a new seqScan plan for it etc. (And building the remote query too). 

What do you think?

jasonmp85:
Another way of looking at it is that the multi-shard logic _doesn't care_ how many shards are in the list. It can give the same answers as the single-shard mode, it just gets there another way (which is why I changed it to `useMultiShardSelectLogic`, i.e. it's just a logic modality, not an either/or.

As to the questions:
  1. Not sure, I'll check
  2. I'll see how much code it adds. I didn't want to add extra code for this one (rather rare) case. It doesn't seem worth the extra complication to go to ternary logic (zero, one, or more) rather than simple binary, and the performance isn't really an issue either (it's indistinguishable from a local query, IMO).

I had another thought. The query in question is actually not satisfiable. Does PostgreSQL have special logic in this case? It seems like it could just short-circuit (not even grab locks on relations) and return zero rows…

jasonmp85:
Tasks for me:

  - [ ] See how single-shard logic handles this if I remove the `ereport` call
  - [ ] Investigate whether PostgreSQL has short-circuit logic for unsatisfiable queries
  - [ ] Look into adding entirely new case for zero-shard queries, see how much extra code it adds

sumedhpathak:
I was thinking of two options:
- If 0 shards, mark it as a non-pgshard-plan. This would just hit the empty shell table on the master and return 0 rows. (I don't like it as much to mark it as non-pgshard)
- Mark as pg_shard, create empty task-list distributed plan, and in the executor just defer to the standard_executor. This reduces code, with the assumption that the shell table on the master is empty. The right solution would be to return 0 rows, but we could comment it accordingly.

On the single-shard piece, its not important. I was just wondering if we chose to combine this with multi-shard logic because single-shard logic doesn't handle it for some reason.

jasonmp85:
Replacing the `ereport` with a `return` yields some weird results. The query completes, but it doesn't look like a normal `SELECT`. Rather than showing an empty result with `(0 rows)` it shows `SELECT 0`, which is a result more akin to what I'd see from an `INSERT`. This is probably because the code no longer starts up the destination, etc.

jasonmp85:
As far as built-in support for unsatisfiable queries, PostgreSQL seems to have certain code that determines unsatisfiable combinations of index conditions, but I'm not seeing anything we can just pick-up and use. Besides, if there is an aggregate over the scan, just returning zero rows is wrong (for instance, a `count(*)` over zero shards is a row with `0`, not the empty set).

pg_shard.c
```diff
@@ -1107,6 +1105,17 @@ PgShardExecutorStart(QueryDesc *queryDesc, int eflags)
...
 			EState *executorState = NULL;
+			bool zeroShardQuery = list_length(distributedPlan->taskList) == 0;
+
+			/* if query involves zero shards, just let it hit local table */
+			if (zeroShardQuery)
+			{
+				Plan *originalPlan = distributedPlan->originalPlan;
+				plannedStatement->planTree = originalPlan;
+
+				PgShardExecutorStart(queryDesc, eflags);
```
sumedhpathak: Hmm. Instead of PgShardExecutorStart calling itself (and then returning), can't we have a check and then fall to calling the hooks explicitly?
We already have a selectFromMultipleShards statement, so I was thinking of something like:
```
if (zeroShardQuery)
{
   // Comment describing why we do this, and why it works.
  // replace distributed plan with original plan 
  if (hook) ... (call hook)
   else standard_ExecutorStart();
}
else if (multiShardQuery)
{

}
else // singleShardQuery
{
 
}
```

This is then more readable, and I don't have to figure out after-effects of calling the function again.

jasonmp85: Sorry, the DRY principle has a pathological grip on me. I'll move the "decide which hook to run" into its own function, though… we'll have it in three places in this function otherwise (copy-pasting it again was why I went with the recursive call).

I think I can rework this the way you asked without adding too many lines of code.

jasonmp85:
@sumedhpathak this has been updated. I see now that `NextExecutorStartHook` might be a bit confusing in the context of `PreviousExecutorStartHook`, but tell me whether the code itself looks OK and then I can think up a new name.

pg_shard.c
```diff
@@ -610,9 +611,10 @@ ExtractRangeTableEntryWalker(Node *node, List **rangeTableList)
 
 /*
  * DistributedQueryShardList prunes the shards for the table in the query based
- * on the query's restriction qualifiers, and returns this list. If the function
- * cannot find any shards for the distributed table, it errors out. In other sense,
- * the function errors out or returns a non-empty list.
+ * on the query's restriction qualifiers, and returns this list. It is possible
+ * that all shards will be pruned if a query's restrictions are unsatisfiable.
+ * In that case, this function can return an empty list; however, if the table
+ * being queried has no shards created whatsoever, this function errors out.
```
sumedhpathak: Do we have a regression test for the case when there are no shards in the table at all? If so, then you are good to :shipit:

jasonmp85: I'll check. Also, this change works for `DELETE` and `UPDATE`, too, so I'll add tests for those (zero shard insert is AFAIK impossible, unless a shard is missing outright).

jasonmp85: The case you mentioned was addressed in #38 and I've added the tests I mentioned.

PR #71, opened 2015-02-05
Add PGXN metadata

### LINKED ISSUE #67: Add package to PostgreSQL package managers
The obvious one is [PGXN](http://pgxn.org/), though [pex](https://github.com/petere/pex) has caught my attention as a more familiar alternative (it's similar to Homebrew, from what I can tell).

jasonmp85:
So we can upload this package to PGXN.

resolves #67

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### pg_shard v1.0.2 (February 24, 2015) ###
+
+* Adds META.json file for PGXN
+
 ### pg_shard v1.0.1 (December 4, 2014) ###
 
 * Minor documentation fixes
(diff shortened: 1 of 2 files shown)


pykello:
This looks good and matches the [specification](http://pgxn.org/spec/) for the most part. The only thing that is a bit different from the examples in spec is the name/email pairs. The [examples in spec](http://pgxn.org/spec/#maintainer) suggest something like ```"Jason Petersen <jason@citusdata.com>"``` instead of ```"\"Jason Petersen\" <jason@citusdata.com>"```.

I don't have comments on the other parts, and I think this can be merged.

jasonmp85:
@pykello I looked at the relevant RFCs for email addresses, and they require quotes around the name part of an address if it has any whitespace characters. Because these RFCs are _very_ hard to read and fairly obscure, it didn't surprise me their example was wrong. But for best compliance we want the quotes.

PR #72, opened 2015-02-05
Bump default version; add upgrade script

### LINKED ISSUE #69: Add v1.1 upgrade script
We added a new user-facing function in #56 but did so directly in the main `pg_shard` SQL file. We need to bump our version number and add an upgrade script. I'm thinking during develop the script (and version) should be named `develop` or something and we can fill it in with a real version number during release. This will always give us a placeholder to add new functions during development and then we can "freeze" that version during release by renaming the file to the actual version being released.

jasonmp85:
Adding a new UDF means adding an upgrade script to augment existing installations of pg_shard.

fixes #69

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ else
 endif
 
 EXTENSION = pg_shard
-DATA = pg_shard--1.0.sql
+DATA = pg_shard--1.1.sql pg_shard--1.0--1.1.sql
 
 # Default to 5432 if PGPORT is undefined. Replace placeholders in our tests
 # with actual port number in order to anticipate correct output during tests.
--- a/pg_shard--1.0--1.1.sql
+++ b/pg_shard--1.0--1.1.sql
@@ -1,0 +1,62 @@
+CREATE FUNCTION partition_column_to_node_string(table_oid oid)
+RETURNS text
+AS 'MODULE_PATHNAME'
+LANGUAGE C;
+
+COMMENT ON FUNCTION partition_column_to_node_string(oid)
+		IS 'return textual form of distributed table''s partition column';
+
+CREATE FUNCTION sync_table_metadata_to_citus(table_name text) RETURNS VOID
+AS $sync_table_metadata_to_citus$
+	DECLARE
+		table_relation_id CONSTANT oid NOT NULL := table_name::regclass::oid;
+		dummy_shard_length CONSTANT bigint := 0;
+	BEGIN
+		-- copy shard placement metadata
+		INSERT INTO pg_dist_shard_placement
+					(shardid,
+					 shardstate,
+					 shardlength,
+					 nodename,
+					 nodeport)
+		SELECT shard_id,
+			   shard_state,
+			   dummy_shard_length,
+			   node_name,
+			   node_port
+		FROM   pgs_distribution_metadata.shard_placement
+		WHERE  shard_id IN (SELECT id
+							FROM   pgs_distribution_metadata.shard
+							WHERE  relation_id = table_relation_id);
+
+		-- copy shard metadata
+		INSERT INTO pg_dist_shard
+					(shardid,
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -25,7 +25,7 @@ else
 endif
 
 EXTENSION = pg_shard
-DATA = pg_shard--1.0.sql
+DATA = pg_shard--1.1.sql pg_shard--1.0--1.1.sql
```
sumedhpathak: So how does this work? Will both files be executed on a ``CREATE EXTENSION`` command?

As I understand, and ``ALTER EXTENSION UPDATE`` on a 1.0 installation will add the new UDF.

Should we be moving the UDF to the 1.1 file, and keeping the above line as pg_shard--1.1.sql only? If I run a ``CREATE EXTENSION``, shouldn't I expect the 1.1 file to include everything?
I'm not sure, so I'm just asking.

jasonmp85: No, the extension will come with exactly one version: 1.1. If you want 1.0 you'd download that version instead.

So for installing from scratch, the 1.1 script is used. For upgrading from 1.0, PostgreSQL will route a path from 1.0 to 1.1. Here it's trivial (a single file), so it'll execute that to perform the upgrade.

jasonmp85:
As a sanity check, I looked over at [cstore](https://github.com/citusdata/cstore_fdw), which looks exactly like pg_shard will after this Pull Request.

Makefile
sumedhpathak: > The UDF is already in the 1.1 file. We checked it into 1.0, but this Pull Request renames the file to 1.1. The 1.1 standalone file should include everything (double check for me).

Ah, ok. That part I didn't realize. For some reason, I thought the UDF was only in the 1.0-1.1.sql file. The rest of it makes sense then. I checked, and the 1.1 sql file looks good.

sumedhpathak: One last question then. The create-extension command also supports a version. If I specify create-extension version 1.0, would that work? (And create 1.0 without the UDF?)

After checking on that, and assuming we've tested 1.0-1.1 upgrade and a 1.1 fresh install, you're good to :shipit:

jasonmp85: The only way to install version 1.0 would be to download 1.0 and use that. The state of affairs for our extensions seems to be:

  * The latest download contains a script to install that latest version
  * It also contains upgrade scripts from earlier versions

jasonmp85:
Testing (not currently automated in any fashion):

  * Checked out tag `v1.0.2` and ran `make uninstall; make install`. This places the control and 1.0 SQL file in my PostgreSQL extension area
  * Checked out this branch and ran `make uninstall; make install`. This replaces the control file with one defining `1.1` as the default version and adds the `1.1` upgrade (and install) script. This does _not_ remove the `1.0` install script added by the first step
  * Modified the `init.sql` in our tests to first install 1.0 and then upgrade to the latest known version
  * Ran tests and verified they passed

Here is the SQL for the upgrade:

```sql
CREATE EXTENSION pg_shard VERSION '1.0';
ALTER EXTENSION pg_shard UPDATE;
# \dx
#                                List of installed extensions
#    Name   | Version |   Schema   |                       Description                       
# ----------+---------+------------+---------------------------------------------------------
#  pg_shard | 1.1     | public     | extension for sharding across remote PostgreSQL servers
#  plpgsql  | 1.0     | pg_catalog | PL/pgSQL procedural language
# (2 rows)
```

jasonmp85:
So yeah, the two caveats are:

  * Future versions don't remove the "main" install script from previous versions
  * Our CI system (Travis) isn't currently testing upgrade paths (these could become numerous as time goes on anyhow, so we'll have to see how much support we need for previous versions before we go down the road of automating this).

But this patch works, works with an upgrade from 1.0, and matches what we're doing in cstore.

PR #73, opened 2015-02-06
Feature issue#16

### LINKED ISSUE #16: Detect unsupported partition key types during dist. table creation
If a type lacks a hash function, we should not allow tables to be distributed using it as the type of their partition key. Right now, doing so results in the cryptic `cache lookup failed for function 0` during `INSERT` commands.

`master_create_distributed_table` should detect these types, refuse to create a distributed table, and print a helpful message.

onderkalaci:
This pull request aims to prevent tables to be distributed with columns that cannot be used as distribution column.

fixes #16

**Code Review Tasks**

  - [x] Remove `get_attnum` section and replace with call to `ColumnNameToColumn`
  - [x] Rewrite `GetSupportRoutine` as `SupportFunctionForColumn` and make it take a `Var *`
  - [x] Move check for default operator class into `SupportFunctionForColumn`
  - [x] Update error codes, messages, and codes as specified
  - [x] Reduce variable scope as much as possible
  - [x] Fix SQL formatting issues in test

### DIFF
--- a/create_shards.c
+++ b/create_shards.c
@@ -56,7 +56,7 @@ static List * ParseWorkerNodeFile(char *workerNodeFilename);
 static int CompareWorkerNodes(const void *leftElement, const void *rightElement);
 static bool ExecuteRemoteCommand(PGconn *connection, const char *sqlCommand);
 static text * IntegerToText(int32 value);
-static Oid GetSupportRoutine(Oid columnOid, Oid accessMethodId, int16 supportFunctionNumber);
+static Oid GetSupportFunction(Oid columnOid, Oid accessMethodId, int16 supportFunctionNumber);
 
 
 /* declarations for dynamic loading */
@@ -83,8 +83,8 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 	Var* partitionColumnVar = NULL;
 	Oid partitionColumnTypeId = InvalidOid;
 	Oid partitionColumnOpClassId = InvalidOid;
-	Oid btreeSupportRoutine = InvalidOid;
-	Oid hashSupportRoutine = InvalidOid;
+	Oid btreeSupportFunction = InvalidOid;
+	Oid hashSupportFunction = InvalidOid;
 
 	/* verify target relation is either regular or foreign table */
 	relationKind = get_rel_relkind(distributedTableId);
@@ -134,9 +134,9 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 	/* depending on the partition type, check for the existence of support procedure */
 	if (partitionMethod == HASH_PARTITION_TYPE)
 	{
-		hashSupportRoutine = GetSupportRoutine(partitionColumnTypeId, HASH_AM_OID,
-											   HASHPROC);
-		if (hashSupportRoutine == InvalidOid)
+		hashSupportFunction = GetSupportFunction(partitionColumnTypeId, HASH_AM_OID,
+											     HASHPROC);
+		if (hashSupportFunction == InvalidOid)
 		{
 			ereport(ERROR, (errcode(ERRCODE_WRONG_OBJECT_TYPE),
 							errmsg("cannot distribute relation: \"%s\"", tableName),
@@ -153,9 +153,9 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 		 * TODO: Add regression tests for this check when RANGE_PARTITION_TYPE is
 		 * supported.
 		 */
-		btreeSupportRoutine = GetSupportRoutine(partitionColumnTypeId, BTREE_AM_OID,
-												BTORDER_PROC);
-		if (btreeSupportRoutine == InvalidOid)
+		btreeSupportFunction = GetSupportFunction(partitionColumnTypeId, BTREE_AM_OID,
+												  BTORDER_PROC);
+		if (btreeSupportFunction == InvalidOid)
 		{
 			ereport(ERROR, (errcode(ERRCODE_WRONG_OBJECT_TYPE),
 							errmsg("cannot distribute relation: \"%s\"", tableName),
@@ -606,16 +606,16 @@ IntegerToText(int32 value)
 
 
 /*
- *	GetSupportRoutine helps to find the support routine given a column, an access method
+ *	GetSupportFunction helps to find the support function given a column, an access method
  *	and id of a support function. This function returns InvalidOid if there is no support
- *	routine associated with the data type of the column or there is no default operator
+ *	function associated with the data type of the column or there is no default operator
  *	class for the data type of the column.
  */
 Oid
-GetSupportRoutine(Oid columnOid, Oid accessMethodId, int16 supportFunctionNumber)
+GetSupportFunction(Oid columnOid, Oid accessMethodId, int16 supportFunctionNumber)
 {
 	Oid operatorFamilyId = InvalidOid;
-	Oid supportRoutineOid = InvalidOid;
+	Oid supportFunctionOid = InvalidOid;
 	Oid operatorClassId = GetDefaultOpClass(columnOid, accessMethodId);
 
 	if (operatorClassId == InvalidOid)
@@ -625,8 +625,8 @@ GetSupportRoutine(Oid columnOid, Oid accessMethodId, int16 supportFunctionNumber
 	}
 
 	operatorFamilyId = get_opclass_family(operatorClassId);
-	supportRoutineOid = get_opfamily_proc(operatorFamilyId, columnOid, columnOid,
-										  supportFunctionNumber);
+	supportFunctionOid = get_opfamily_proc(operatorFamilyId, columnOid, columnOid,
+										   supportFunctionNumber);
 
-	return supportRoutineOid;
+	return supportFunctionOid;
 }
--- a/create_shards.c
+++ b/create_shards.c
@@ -80,7 +80,7 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 	char relationKind = '\0';
 	char *partitionColumnName = text_to_cstring(partitionColumnText);
 	char *tableName = text_to_cstring(tableNameText);
-	Var *partitionColumnVar = NULL;
+	Var *partitionColumn = NULL;
 
 	/* verify target relation is either regular or foreign table */
 	relationKind = get_rel_relkind(distributedTableId);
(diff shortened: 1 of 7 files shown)


create_shards.c
```diff
@@ -97,6 +106,35 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
...
+						errhint("Try another column as the distribution column.")));
+	}
+
+	lessEqualOperatorId = GetOperatorByType(partitionColumnTypeId, BTREE_AM_OID,
+											BTLessEqualStrategyNumber);
+	greaterEqualOperatorId =  GetOperatorByType(partitionColumnTypeId, BTREE_AM_OID,
+												BTGreaterEqualStrategyNumber);
+
+	/* if partition column does not support <= and >= operators, error out*/
+	if (lessEqualOperatorId == InvalidOid ||  greaterEqualOperatorId == InvalidOid)
```
onderkalaci: I could not find any type, which has non-InvalidOid partitionColumnOpClassId, with invalid equality operators.
But, to be on the safe side, I added this extra control.

jasonmp85: I believe this check only makes sense for tables that use range partitioning: for hash partitioning, the shards are always defined using 32-bit integers, and we can always find the inequality operator for that.

If using range partitioning, shard boundaries _will_ be defined using the partition column type, so it makes sense in that case to check for inequality operators.

Can you change this section to do something like:

```
if using hash partitioning
  check for presence of hash operator for partition column type
else if using range partitioning
  check for existence of inequality operators for partition column type
else
  error out that partition type is unexpected
```

jasonmp85:
Hey @onderkalaci — I know I said I'd email @sumedhpathak and @ozgune about breaking the error checking out into its own function, but what I hadn't realized is that `master_create_distributed_table` is _all_ error checking! The GitHub view cut off right after the call to `InsertPartitionRow`, so I didn't see that we return immediately after.

So there is no argument to support breaking error handling into its own function here: it makes sense that `master_create_distributed_table` is all error handling code since the logic is so simple at this point (a single function call).

Which just leaves my point about the inequality operator checks. I'll see whether there are any other stylistic things that need addressing and add a checklist at the top of this review. After those changes I think we're good to merge.

onderkalaci:
Hey @jasonmp85, @sumedhpathak 

Below I summerize what is included in the updated pull request:

* A new function "*GetSupportRoutine()*" is added. Basically, aim of this function is to find the support routine wrt partition method. 

* For instance, this function helps us to check if the partition column has a hash function if hash partitioning is used. Similarly, check if partition column has required comparison operator if range partitioning is used.  This function is compatible with citus codes.

* Two functions changed from "static" to "extern" which are used in the pull request:
	1) distribution_metadata.c @*ColumnNameToColumn(Oid relationId, char *columnName);*
	2) pruneShardList.c @*GetOperatorByType(Oid typeId, Oid accessMethodId, int16 strategyNumber);*

* Two checks are added to prevent table to be distributed:
	1) If partition column data type does not have a default operator class, error out (ie:json data type).
	2) Depending on the partition type, check for the existence of support procedure. For details, see below bullet item.

* Added two tests to increase the test coverage. One for each item above. In order to add the test, I created a type, function, operator, operator class and operator family. The reason that I had to create them is that I could not find any internal data type that does not have support routine.

* One controversial issue in the code is that, I added check for support routine for range partitioning.
However, that part of the code is not possible to be executed because we reject range partitioning before that code is executed. So, I added a comment saying that when we support range partitioning, we should add regression test. Should I delete or handle this case in another way? Or it is OK?

* One future issue that we need to do is to apply the composite type bug changes to pg_shard, which is on CR in citus currently.

create_shards.c
```diff
@@ -50,6 +56,7 @@ static List * ParseWorkerNodeFile(char *workerNodeFilename);
 static int CompareWorkerNodes(const void *leftElement, const void *rightElement);
 static bool ExecuteRemoteCommand(PGconn *connection, const char *sqlCommand);
 static text * IntegerToText(int32 value);
+static Oid GetSupportRoutine(Oid columnOid, Oid accessMethodId, int16 supportFunctionNumber);
```
jasonmp85: `Routine` is a weird word. I don't see any functions in CitusDB or PostgreSQL that use this term except those related to foreign data wrappers. Most comments refer to this as a _support function_, and I think we should follow that pattern.

create_shards.c
```diff
@@ -91,12 +103,69 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
...
 	}
 
-	/* we only support hash partitioning method for now */
-	if (partitionMethod != HASH_PARTITION_TYPE)
+	partitionColumnVar = ColumnNameToColumn(distributedTableId, partitionColumnName);
+	partitionColumnTypeId = partitionColumnVar->vartype;
+
+	if (partitionMethod == HASH_PARTITION_TYPE)
+	{
+		partitionColumnOpClassId = GetDefaultOpClass(partitionColumnTypeId, HASH_AM_OID);
```
jasonmp85: Calling `GetDefaultOpClass` here and doing validity checking further down is confusing to me. At first I was going to suggest moving the validity checking into this `if` block, but I really think we should just move the validity checking into the `GetSupportRoutine` function, i.e. have it `ereport` directly rather than returning `InvalidOid` if there is no default op class for the specified type.

create_shards.c
```diff
@@ -91,12 +103,69 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 		ereport(ERROR, (errmsg("could not find column: %s", partitionColumnName)));
 	}
 
-	/* we only support hash partitioning method for now */
-	if (partitionMethod != HASH_PARTITION_TYPE)
+	partitionColumnVar = ColumnNameToColumn(distributedTableId, partitionColumnName);
```
jasonmp85: Move the call to `ColumnNameToColumn` directly into the declaration of the variable, at the top of this function. Similarly, move the definition of `partitionColumnTypeId` up there.

create_shards.c
```diff
@@ -73,6 +80,11 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 	AttrNumber partitionColumnId = InvalidAttrNumber;
 	char *partitionColumnName = text_to_cstring(partitionColumnText);
 	char *tableName = text_to_cstring(tableNameText);
+	Var* partitionColumnVar = NULL;
+	Oid partitionColumnTypeId = InvalidOid;
+	Oid partitionColumnOpClassId = InvalidOid;
```
jasonmp85: A `partitionColumnOpClassId` variable is no longer necessary if we move the `ereport` call into `GetSupportRoutine`, so we should kill it entirely. Additionally, the `btreeSupportRoutine` and `hashSupportRoutine` variables should have their scope reduced by moving their declaration into the relevant blocks below.

create_shards.c
```diff
@@ -91,12 +103,69 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
...
+	partitionColumnTypeId = partitionColumnVar->vartype;
+
+	if (partitionMethod == HASH_PARTITION_TYPE)
+	{
+		partitionColumnOpClassId = GetDefaultOpClass(partitionColumnTypeId, HASH_AM_OID);
+	}
+	else
 	{
+		/* we only support hash partitioning right now */
 		ereport(ERROR, (errmsg("unsupported partition method: %c", partitionMethod)));
```
jasonmp85: Since you're modifying this function, can you update this `ereport` call to use the `ERRCODE_FEATURE_NOT_SUPPORTED` status code? It's most appropriate here.

I like the error message exactly as it stands, though we should add an error detail: _pg_shard only supports hash partitioning._

sql/create_shards.sql.tmpl
```diff
@@ -13,12 +13,38 @@ CREATE FUNCTION create_table_then_fail(cstring, integer)
...
+CREATE FUNCTION dummy_type_function(dummy_type, dummy_type) RETURNS boolean
+AS ' SELECT TRUE;'
+LANGUAGE SQL
+IMMUTABLE
+RETURNS NULL ON NULL INPUT;
+
+CREATE OPERATOR = (
+    LEFTARG = dummy_type,
+    RIGHTARG = dummy_type,
+    PROCEDURE = dummy_type_function);
```
jasonmp85: Can you put the closing parenthesis on the next line (as you did with `CREATE TYPE`)?

sql/create_shards.sql.tmpl
```diff
@@ -13,12 +13,38 @@ CREATE FUNCTION create_table_then_fail(cstring, integer)
...
 
 -- ===================================================================
+-- create test type, operator, operator family and operator class
+-- ===================================================================
+CREATE TYPE dummy_type AS (
+    i integer
+);
+
+CREATE FUNCTION dummy_type_function(dummy_type, dummy_type) RETURNS boolean
+AS ' SELECT TRUE;'
```
jasonmp85: Remove the space before `SELECT` and the open quote.

create_shards.c
```diff
@@ -534,3 +603,30 @@ IntegerToText(int32 value)
 
 	return valueText;
 }
+
+
+/*
+ *	GetSupportRoutine helps to find the support routine given a column, an access method
```
jasonmp85: `GetSupportRoutine` is a bad name because _routine_ isn't a word PostgreSQL uses frequently. I see _support function_ and _support procedure_, though the former seems more frequent, so let's go with that…

`GetSupportFunction` should be the name of this function.

create_shards.c
```diff
@@ -534,3 +603,30 @@ IntegerToText(int32 value)
...
+ *	and id of a support function. This function returns InvalidOid if there is no support
+ *	routine associated with the data type of the column or there is no default operator
+ *	class for the data type of the column.
+ */
+Oid
+GetSupportRoutine(Oid columnOid, Oid accessMethodId, int16 supportFunctionNumber)
```
jasonmp85: At first I was going to say `columnOid` is misleading because this function doesn't have to know that the `typeId` it receives belongs to a column, but seeing that here gave me an idea…

jasonmp85:
Ok, so after staring at this for a while, I think I know what I'd like to see happen…

In some code you haven't modified, there's a section like:

```c
partitionColumnId = get_attnum(distributedTableId, partitionColumnName);
if (partitionColumnId == InvalidAttrNumber)
{
	ereport(ERROR, (errmsg("could not find column: %s", partitionColumnName)));
}
```

This error checking is duplicated by `ColumnNameToColumn`, which you call later. I think the above section should be _removed entirely_ and replaced by the call to `ColumnNameToColumn` (step one).

Then, `GetSupportRoutine` should be renamed `SupportFunctionForColumn` and it should be modified to accept that `Var *` (step two).

If the type does not have a default operator class, `SupportFunctionForColumn` should call `ereport` itself with a suitable error message (step three).

I think this could help code flow quite a bit while cutting down on verbosity and duplication.

create_shards.c
```diff
@@ -73,6 +80,11 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 	AttrNumber partitionColumnId = InvalidAttrNumber;
 	char *partitionColumnName = text_to_cstring(partitionColumnText);
 	char *tableName = text_to_cstring(tableNameText);
+	Var* partitionColumnVar = NULL;
+	Oid partitionColumnTypeId = InvalidOid;
+	Oid partitionColumnOpClassId = InvalidOid;
+	Oid btreeSupportRoutine = InvalidOid;
```
jasonmp85: Move the declaration of `btreeSupportRoutine` and `hashSupportRoutine` into the respective `if`/`else if` blocks to reduce their scope… no reason to declare them up here. And when you move them, you can combine the declaration and assignment into a one-line definition for each.

create_shards.c
```diff
@@ -91,12 +103,69 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
...
 	}
 
+	/*
+	 * If data type of the partition column does not have a default operator class, we
+	 * should prevent distribution of the table on this column. It is because there are
+	 * no operators/procedures that can manipulate the value of the column.
+	 */
+	if (partitionColumnOpClassId == InvalidOid)
+	{
+		ereport(ERROR, (errcode(ERRCODE_WRONG_OBJECT_TYPE),
```
jasonmp85: `ERRCODE_WRONG_OBJECT_TYPE` isn't appropriate here. Let's stick with what other functions do when they get an `InvalidOid` here. I think the following should work:

```c
ereport(ERROR,
		(errcode(ERRCODE_UNDEFINED_OBJECT),
		 errmsg("data type %s has no default operator class for specified partition method",
		 		format_type_be(attrType)),
		 errdetail("Partition column types must have a default operator class defined.")));
```

create_shards.c
```diff
@@ -91,12 +103,69 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
...
+	}
+
+	/* depending on the partition type, check for the existence of support procedure */
+	if (partitionMethod == HASH_PARTITION_TYPE)
+	{
+		hashSupportRoutine = GetSupportRoutine(partitionColumnTypeId, HASH_AM_OID,
+											   HASHPROC);
+		if (hashSupportRoutine == InvalidOid)
+		{
+			ereport(ERROR, (errcode(ERRCODE_WRONG_OBJECT_TYPE),
```
jasonmp85: Again, `ERRCODE_WRONG_OBJECT_TYPE` is inappropriate. `arrayfuncs.c` has a _ton_ of examples of errors to use when a specific function lookup fails. Something like:

```c
ereport(ERROR,
		(errcode(ERRCODE_UNDEFINED_FUNCTION),
		 errmsg("could not identify a comparison function for type %s",
		 		format_type_be(element_type))
		 errdetail("Partition column types must have a comparison function defined to use range partitioning.")));
```

jasonmp85:
Ok @onderkalaci — I've left my checklist at the top of this PR. I think the reorganization is enough that I'll want to check this out once more before a merge, but assuming it's mostly in shape by morning (MDT) I'll merge it myself if it looks good.

onderkalaci:
@jasonmp85 Thanks for the review. 
I think that now code looks better than previous implementation.

jasonmp85:
@onderkalaci By default, Coveralls will complain if coverage decreases. We discussed the coverage of your patch (and how it couldn't be complete until we support range partitioning) and I expected coverage to decrease slightly.

If you notice, the merge button is still present, and _nothing_ can ever stop any of us from merging from the command line. So it's just a slight inconvenience designed to make sure you are aware you lowered coverage.

In this case, it's expected, so no biggie.

create_shards.c
```diff
@@ -84,17 +92,53 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
...
-	if (partitionMethod != HASH_PARTITION_TYPE)
+	/* depending on the partition type, check for the existence of support function */
+	if (partitionMethod == HASH_PARTITION_TYPE)
 	{
-		ereport(ERROR, (errmsg("unsupported partition method: %c", partitionMethod)));
+		Oid hashSupportFunction = SupportFunctionForColumn(partitionColumnVar,
+														   HASH_AM_OID, HASHPROC);
+		if (hashSupportFunction == InvalidOid)
+		{
+			Oid partitionColumnTypeId = partitionColumnVar->vartype;
```
jasonmp85: I don't see the point in this variable existing, so I'm going to remove it before merging (it's only referenced once).

create_shards.c
```diff
@@ -534,3 +578,40 @@ IntegerToText(int32 value)
...
+	/*
+	 * If data type of the partition column does not have a default operator class,
+	 * we should not continue to get support function. A type with no default op
+	 * class cannot have any support functions.
+	 */
+	if (operatorClassId == InvalidOid)
+	{
+		ereport(ERROR, (errcode(ERRCODE_UNDEFINED_OBJECT),
+						errmsg("data type %s has no default operator class for specified"
+								" partition method", format_type_be(columnOid)),
```
jasonmp85: Slight alignment issue here. I'll fix before merge.

prune_shard_list.c
```diff
@@ -305,7 +304,7 @@ LookupOperatorByType(Oid typeId, Oid accessMethodId, int16 strategyNumber)
  * GetOperatorByType returns the operator oid for the given type, access
  * method, and strategy number.
  */
-static Oid
+Oid
```
jasonmp85: I think we're starting to see we might want to move some of these common functions out into their own file.

create_shards.c
```diff
@@ -534,3 +578,40 @@ IntegerToText(int32 value)
...
+Oid
+SupportFunctionForColumn(Var *partitionColumn, Oid accessMethodId,
+		                 int16 supportFunctionNumber)
+{
+	Oid operatorFamilyId = InvalidOid;
+	Oid supportFunctionOid = InvalidOid;
+	Oid columnOid = partitionColumn->vartype;
+	Oid operatorClassId = GetDefaultOpClass(columnOid, accessMethodId);
+
+	/*
```
jasonmp85: This comment is actually incorrect, I think… [This documentation section](http://www.postgresql.org/docs/9.4/static/indexes-opclass.html) shows that an index can be defined using a customer operator class rather than the default. So it's possible we could support types without default operator classes, so long as the user specified which custom operator class to use at the time of distribution. I'll update this comment to clarify that we just don't support anything other than using a default operator class at the moment.

jasonmp85:
I've tested this on the latest CitusDB, PostgreSQL 9.3, and PostgreSQL 9.4 on OS X. Travis'll take care of Linux. After it passes I'm gonna merge.

PR #76, opened 2015-02-23
Feature issue#49

### LINKED ISSUE #49: Prevent deletion of metadata when still in use

onderkalaci:
This pull request includes commits which add dependency control for pg_shard extension and distributed tables.

Fixes #49 

**Code Review Tasks**

  - [x] Rebase this branch onto `develop` (i.e. `git fetch; git rebase -i origin/develop`) or merge develop into this branch
  - [x] Remove `DROP TABLE` logic and all functions/tests used by that logic
  - [x] Tell me what happens when you run `DROP EXTENSION pg_shard` against a database where `pg_shard` isn't yet created (i.e. does your block run, or is it skipped?)
  - [x] Don't do `if (booleanVar == true)`, just do `if (booleanVar)`
  - [x] Always use the bounded version of string/memory functions (`strncmp`, never `strcmp`)
  - [x] Pull up `DistributedTablesExist` so it's not called in a loop
  - [x] Don't check a condition just to set a boolean flag to true in an if block. Just set the boolean flag to the condition's value


pg_shard.c
```diff
@@ -2007,6 +2008,107 @@ PgShardProcessUtility(Node *parsetree, const char *queryString,
 			}
 		}
 	}
+	else if (statementType == T_DropStmt)
+	{
+		DropStmt *dropStatement = (DropStmt *) parsetree;
```
jasonmp85: The nesting gets pretty deep here. Can you move the drop logic into its own function?

pg_shard.c
```diff
@@ -2007,6 +2008,107 @@ PgShardProcessUtility(Node *parsetree, const char *queryString,
...
+		bool dropBehavior = dropStatement->behavior;
+		bool extensionInstalled = false;
```
jasonmp85: `bool extensionInstalled = get_extension_oid(PG_SHARD_EXTENSION_NAME, missingOK) != InvalidOid;`

pg_shard.c
```diff
@@ -2007,6 +2008,107 @@ PgShardProcessUtility(Node *parsetree, const char *queryString,
...
+		Oid extensionOid = InvalidOid;
+		bool missingOK = true;
+
+		extensionOid = get_extension_oid(PG_SHARD_EXTENSION_NAME, missingOK);
```
jasonmp85: Get rid of `extensionOid` entirely and just do this here:

```c
extensionInstalled = (get_extension_oid(PG_SHARD_EXTENSION_NAME, missingOK) != invalidOid);
```

pg_shard.c
```diff
@@ -2007,6 +2008,107 @@ PgShardProcessUtility(Node *parsetree, const char *queryString,
...
+		if (extensionOid != InvalidOid)
```
jasonmp85: This `if` statement is kind of pathological. Just do:

```c
extensionInstalled = (extensionOid != InvalidOid);
```

pg_shard.c
```diff
@@ -2007,6 +2008,107 @@ PgShardProcessUtility(Node *parsetree, const char *queryString,
...
+			{
+				List *tableNameList = (List *) lfirst(dropObjectCell);
+				RangeVar *rangeVar = makeRangeVarFromNameList(tableNameList);
+				Oid relationId = RangeVarGetRelid(rangeVar, AccessShareLock, true);
+
+				/*
+				 * If a distributed table is dropped with CASCADE modifier, first inform,
+				 * then delete all the dependent metadata information.
+				 */
+				if (extensionInstalled == true && IsDistributedTable(relationId))
```
jasonmp85: The `== true` is redundant.

pg_shard.c
```diff
@@ -2007,6 +2008,107 @@ PgShardProcessUtility(Node *parsetree, const char *queryString,
...
+		else if (dropStatement->removeType == OBJECT_EXTENSION)
+		{
+			ListCell *dropStatementObject = NULL;
+
+			foreach(dropStatementObject, dropStatement->objects)
+			{
+				List *objectNameList = lfirst(dropStatementObject);
+				char *objectNameStr = NameListToString(objectNameList);
+
+				if (strcmp(PG_SHARD_EXTENSION_NAME, objectNameStr) == 0)
```
jasonmp85: Use `strncmp` here with `NAMEDATALEN` as the length argument.

pg_shard.c
```diff
@@ -2007,6 +2008,107 @@ PgShardProcessUtility(Node *parsetree, const char *queryString,
...
+					/*
+					 * If pg_shard is dropped with CASCADE modifier, let it to be dropped.
+					 * However, inform the user that shards on the worker nodes will not
+					 * be dropeed.
+					 *
+					 * If CASCADE modifier is not used, check for existence of any
+					 * distributed tables. If there exists any distributed table, do not
+					 * let pg_shard to be dropped.
+					 */
+					if (extensionInstalled == true && dropBehavior == DROP_CASCADE)
```
jasonmp85: `== true` is redundant.

pg_shard.c
```diff
@@ -2007,6 +2008,107 @@ PgShardProcessUtility(Node *parsetree, const char *queryString,
...
+					{
+						ereport(NOTICE, (errmsg("dropping all the metadata tables"),
+										 errdetail("Shards created by the extension are"
+												   " not removed.")));
+					}
+					else
+					{
+						bool distributedTableExists = DistributedTablesExist();
+						if (distributedTableExists == true)
```
jasonmp85: Again, `== true` is pretty redundant for a `bool`.

pg_shard.c
```diff
@@ -2007,6 +2008,107 @@ PgShardProcessUtility(Node *parsetree, const char *queryString,
...
+					 */
+					if (extensionInstalled == true && dropBehavior == DROP_CASCADE)
+					{
+						ereport(NOTICE, (errmsg("dropping all the metadata tables"),
+										 errdetail("Shards created by the extension are"
+												   " not removed.")));
+					}
+					else
+					{
+						bool distributedTableExists = DistributedTablesExist();
```
jasonmp85: Variable should probably be `distributedTablesExist`.

pg_shard.c
jasonmp85: Also, why do we need to check whether the extension is installed at all? Is it possible for the execution to get this far if it's not (i.e. would our utility hook run before PostgreSQL even verifies whether ever object mentioned in the statement exists?)

pg_shard.c
jasonmp85: And should we be calling `DistributedTablesExist` in a loop? Why not call it once at the top of this function and just save the value in a variable?

jasonmp85:
Left some comments, mostly minor for now. Still want to think overnight about this approach, as it's quite a bit of code for something that's conceptually pretty simple. Will add more comments tomorrow morning.

pg_shard.c
onderkalaci: @jasonmp85 I think we need to discuss it .
When I check *DeterminePlannerType()* function, I see that we check existence of pg_shard. First, it seemed unnecessary for me since we already executing pg_shard's code. However, with further debugging I realized that we never get to *_PG_fini()* function.

Am I missing anything such that debugging *_PG_fini()* is different than other functions or so?

But, since we do not get *_PG_fini()*, I had to add a check similar to *DeterminePlannerType()*.

onderkalaci:
@jasonmp85 Well, in the beginning of our discussions, we thought that we should write C codes for this purpose. If we are going to implement it with C (either event trigger or utility hook), we need to write these codes (at least I couldn't find any other way).

So, we may consider writing a SQL event trigger. 

Or, I am not sure, but if we use referential integrity constraints during metadata table creation, would the code become smaller? For instance, deleting a row from partition column propagates to shard and shard_placements. 

I haven't addressed your feedback, I think it is better to decide high level design first.

onderkalaci:
@jasonmp85 we have chatted with @sumedhpathak and thought that we could divide the issue into two. One is for DROP EXTENSION and the other is DROP TABLE. In this way, we can merge DROP EXTENSION for v1.1, which is simpler and requires small code changes. 
So, you can also consider this option for the issue.

jasonmp85:
OK, the more I think about it, the more I think @sumedhpathak and you are on to something with the "let's split this in two" idea…

  * The foreign key concept might have some merit and was precisely what I was thinking when considering whether this could be done with a higher level (less LoC) implementation
  * #49 really only discusses removing the extension, anyhow
  * Much of the code here is boilerplate stuff to remove rows. If we just address the extension dropping, that goes away

@onderkalaci Can you rework this to _just_ handle dropping the extension and we'll open an issue for later about dropping tables?

pg_shard.c
jasonmp85: It's possible the extension is _loaded_ (i.e. the hooks are loaded) but not _created_ (i.e. no one ever ran `CREATE EXTENSION pg_shard`). Since planning presupposes the existence of our metadata tables, trying to plan a `pg_shard` query without running `CREATE EXTENSION` first is problematic (i.e. the `pgs_distribution_metadata` schema and tables won't exist).

Since `DistributedTablesExist` makes the same assumption, it would probably result in an error to call it if `CREATE EXTENSION` had never been called; however, I'm wondering whether this block can _even execute_ if the extension hasn't been created, i.e. will PostgreSQL let execution even _get_ this far, or will it do some sanity checking earlier to find that `DROP EXTENSION pg_shard` shouldn't even execute, because `pg_shard` isn't an extant extension?

As far as `pg_fini` is concerned: it is never called. Extensions should implement it for future-proofing, but at the moment only `pg_init` is called (for libraries listed in `shared_preload_libraries`, for explicit `LOAD` calls, or the first time a C-based UDF it called for a given extension). So you don't have to worry about its effects for now.

jasonmp85: Shorter answer: can you call `DROP EXTENSION pg_shard` _before_ ever calling `CREATE EXTENSION` and see whether this block executes at all?

jasonmp85:
OK, I left all my feedback. Can you remove the `DROP TABLE` logic and address feedback?

pg_shard.c
```diff
@@ -2019,3 +2026,79 @@ PgShardProcessUtility(Node *parsetree, const char *queryString,
...
+
+/*
+ * PgShardProcessUtilityDropCommands is a helper function for PgShardProcessUtility.
+ * This function handles DROP utility commands. Currently, only handles DROP EXTENSION
+ * command.
+ */
+static void
+PgShardProcessUtilityDropCommands(Node *parsetree)
+{
+	DropStmt *dropStatement = (DropStmt *) parsetree;
```
jasonmp85: I think the idiom within PostgreSQL is to cast this _before_ calling the function so the function signature can be more specific.

jasonmp85:
@onderkalaci I hope this isn't too presumptuous of me, but I'm making some changes here before merging:

  * The "if there exists" construction is pretty formal, so I rewrote the comments that used it
  * A lot of the comments seemed pretty verbose, so I abbreviated most, leaving block comments only for things that obviously required special explanation
  * I removed any mention of "only handling `DROP EXTENSION`", which won't make sense unless the reader knows we removed the code to handle `DROP TABLE`
  * When I viewed your method in my editor, it had _many_ levels of nesting, requiring long function calls to split parameters across many lines (up to five!). I'll explain the fix to this below
  * I renamed the `PgShardProcessUtilityDropStatement` function to be very clear about its purpose and changed its parameter to a `DropStmt` pointer

The only change in functionality I made is this: the original implementation would have printed a `NOTICE` if the user specified `CASCADE` _even if no distributed tables existed_. This would be misleading. Now I test early whether any distributed tables exist and just exit early it not.

*Reducing Indentation Levels*

If we have a construct like this:

```c
void
SomeFunction(Params *param)
{
    Variable variable = param->field;
    /* ... */

    if (variable == VALUE)
    {
        Variable other = variable * 2;
        /* ... */
    }
    /* no more code in this function after if statement */
}
```

We can remove a level of nesting by doing this instead:

```c
void
SomeFunction(Params *param)
{
    Variable variable = param->field;
    /* ... */

    if (variable != VALUE)
    {
        return;
    }

    /* code that used to be inside if block */
}
```

I'd say use this judiciously, because it can become a tax on the reader, but if you find yourself with five levels of nesting and you're inside some `if` block that could be replaced with a short-circuit (either a call to `ereport(ERROR, ...)` or a `return`), consider it.

jasonmp85:
Oh yeah, and I changed the `ERROR` to use the _precise_ language used by PostgreSQL when you try to drop a table that's dependent on another table.

A final note: You had `errmsg("some message: %s", POUND_DEFINED_STRING_CONSTANT);` in a couple of places. You can just let the compiler concatenate these strings for you: `errmsg("some message: " POUND_DEFINED_STRING_CONSTANT);`

PR #81, opened 2015-03-06
Fix memory leak on large unbounded queries

### LINKED ISSUE #77: Large, unbounded queries cause crashes
In #70, @jberkus says:

> BTW, right now if you try a simple limit query like the above, you get a client termination because it's trying to drag the entire database from all shards into memory:
> ```
> psocial=# select * from members limit 10;
> server closed the connection unexpectedly
>         This probably means the server terminated abnormally
>         before or while processing the request.
> The connection to the server was lost. Attempting reset: Failed.
> Time: 101571.258 ms
> !> 
> !> \q
> ```

`pg_shard` does _not_ try to drag the entire database into memory: it pulls data from each shard into a tuplestore and then loads that into a temporary table (see #32 for some discussion of this). So something else is causing this crash.

Could be related to the data set size in relation to work memory, perhaps. Whatever the case we should figure it out and decide whether to get a hotfix in for this release.

onderkalaci:
This pull request aims to fix the memory leak created by the call to PostgreSQL's BuildTupleFromCStrings function. Temporary memory context is created for each call to BuildTupleFromCStrings and reseted after the call.

Fixes #77

### DIFF
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -1472,11 +1472,9 @@ StoreQueryResult(PGconn *connection, TupleDesc tupleDescriptor,
 		uint32 columnCount = 0;
 		ExecStatusType resultStatus = 0;
 
-
 		PGresult *result = PQgetResult(connection);
 		if (result == NULL)
 		{
-
 			break;
 		}
 
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -75,6 +75,7 @@
 #include "utils/relcache.h"
 #include "utils/snapmgr.h"
 #include "utils/tuplestore.h"
+#include "utils/memutils.h"
 
 
 /* controls use of locks to enforce safe commutativity */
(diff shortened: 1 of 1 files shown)


jasonmp85:
From an email conversation:

> On Mar 6, 2015, at 10:04 AM, Jason Petersen wrote:
>
> can you clarify why you free the tuple even in the area guarded by its own context? Resetting the context clears all memory in it, right? So this is a no-op.

@onderkalaci replied:

> On Mar 6, 2015, at 10:19 AM, Onder Kalaci wrote:
>
> Oh yes that is no-op. Reset already frees it.

I'm going to update the comment slightly to mention I/O functions (another area that leaks like this mentions them as a culprit) and remove the extra call to `heap_freetuple` before merging (i.e. this is a :shipit:).

jasonmp85:
@jberkus — This should address the crash you had if you want to verify. We're going to follow up with the PostgreSQL hackers about memory leaks in I/O functions and `BuildTupleFromCStrings`, but we believe this particular issue is fixed.

PR #82, opened 2015-03-09
Add support for \copy command via script/trigger

### LINKED ISSUE #61: Generalize COPY-to-INSERT trigger
I've written a `COPY` trigger used by some of our customers which turns a `COPY` command into many `INSERT` commands using a temporary table with the same schema as a sharded table.

Unfortunately it's hardcoded to a specific schema. The first step towards users being able to `COPY` to a sharded table is to generalize this script for any table.

jasonmp85:
The script provides an easy way for users to COPY to a distributed table. It accepts a filename and table name, prepares the table for COPY, performs the COPY, then outputs the number of rows copied. Flags are supported to enable various COPY OPTIONS in the underlying SQL statement.

This branch still needs Makefile changes, better documentation, and a unit test, but I wanted to get the code out there to kick off a review. I'll be pushing the remaining changes here as they come up.

Fixes #61

**Code Review Tasks**

  - [x] Switch back to providing sequence to function
  - [x] Make sequence argument optional
  - [x] Add regression test for sequence behavior
  - [x] Switch to distributed table for regression tests
  - [x] Add `COPY` test with partial failure
  - [x] Modify script to use sequence-based UDF
  - [x] Remove as much shell interpolation as possible (favor `psql` variables)
  - [x] Swap order of arguments in `usage` shell function


scripts/copy_to_distributed_table.sh
```diff
@@ -0,0 +1,95 @@
...
+			;;
+		e)
+			escape=`echo ${OPTARG} | sed s/\'/\'\'/g`
+			options="${options}, ESCAPE '${escape}'"
+			;;
+		H)
+			options="${options}, HEADER true"
+			;;
+		h)
+			usage
```
jasonmp85: Just realized I didn't write a `usage` function yet. Fixing in the afternoon.

jasonmp85:
High-level description of the approach: `prepare_distributed_table_for_copy` accepts a table and creates a function specifically for that table based on its columns. Ideally we'd use `INSERT INTO foo VALUES (($1).*) USING NEW` or even `INSERT INTO foo VALUES ($1.bar, $1.baz) USING NEW`, but these types of field accesses aren't permitted by `pg_shard` (at the moment). It rejects them as not being constant expressions.

So the only type of statement that could work is: `INSERT INTO foo VALUES ($1, $2) USING NEW.bar, NEW.baz`. Since the dynamic text of that statement isn't in the `INSERT` string but in the function itself (in the `USING` clause), this necessitates generating a complete function. So that's what we do.

Once the function has been generated for the given table, it is installed as a trigger on a temporary table exactly like the target table. Then any `COPY` executing against the temporary table will be redirected to the distributed table in a manner accepted by `pg_shard`.

sumedhpathak:
> Ideally we'd use INSERT INTO foo VALUES (($1).*) USING NEW or even INSERT INTO foo VALUES ($1.bar, $1.baz) USING NEW, 

Is ``USING`` an INSERT keyword? I don't see it in the Postgres insert documentation?

jasonmp85:
It's part of [the `EXECUTE` statement](http://www.postgresql.org/docs/9.4/static/plpgsql-statements.html#PLPGSQL-STATEMENTS-EXECUTING-DYN) in PL/pgSQL.

pg_shard--1.1.sql
```diff
@@ -140,3 +140,51 @@ $sync_table_metadata_to_citus$ LANGUAGE 'plpgsql';
...
+									$$;
+		table_tmpl CONSTANT text := 'CREATE TEMPORARY TABLE %I (LIKE %s)';
+		trg_tmpl CONSTANT text := $$CREATE TRIGGER copy_to_insert
+									BEFORE INSERT ON %s
+									FOR EACH ROW EXECUTE PROCEDURE pg_temp.copy_to_insert()$$;
+	BEGIN
+		SELECT relname INTO STRICT table_name FROM pg_class WHERE oid = relation;
+		temp_table_name = format('%s_copy_facade', table_name);
+
+		SELECT array_agg(attname) INTO STRICT attr_names FROM pg_attribute WHERE attrelid = relation AND
```
sumedhpathak: Does this mean we only support COPY of all attributes?

jasonmp85:
We don't frequently write this heavily in PL/pgSQL, so here is a brief primer of features used to help in review:

  * Variables must be listed in `DECLARE` blocks at the top of functions (though you can also have nested blocks in the function itself to have limited scope variables)
  * They can be marked constant to prevent reassignment
  * Double dollar signs can be used to delimit strings (and they can have a token within the dollar signs to allow nested quoting; quotes don't end until a matching dollar symbol is found)
  * `format` has `%I` and `%L` specifiers for "identifier" and "literal", respectively. Helps format things that need to appear in an SQL statement as an identifier (`INSERT INTO identifier`) or literal (`VALUES ('literal')`
  * `SELECT INTO variable` allows the result of a `SELECT` to be assigned to a variable. `STRICT` means we expect a single row
  * `||` is string concatenation
  * `EXECUTE` is basically `eval`: it executes a string which may contain positional parameters (`$1`, `$2`, etc.) bound at runtime to values supplied in a `USING` clause.

pg_shard--1.1.sql
jasonmp85: This means the `INSERT` trigger will generate a statement with _all_ user-visible attributes fully specified. If the overarching `COPY` does not specify all columns, one of two things could happen:

  * If `NEW.missing_col` is provided as `NULL`, we'd explicitly include a `NULL` in the `INSERT`. This could be wrong if there is a default value specified
  * If `NEW.missing_col` already replaces unspecified columns with their default values, the trigger would just explicitly specify the default value, which would be correct.

The copy shell script presently doesn't allow omitting columns: it's expected the CSV or whatnot has the same dimensions as the table. So this is kind of a moot point… I'll check about the `NULL` vs default value distinction.

pg_shard--1.1.sql
```diff
@@ -140,3 +140,51 @@ $sync_table_metadata_to_citus$ LANGUAGE 'plpgsql';
...
+		func_tmpl CONSTANT text  := $$  CREATE FUNCTION pg_temp.copy_to_insert() RETURNS trigger
+										AS $cti$
+											BEGIN
+												EXECUTE %L USING %s;
+												PERFORM nextval(%L);
+												RETURN NULL;
+											END;
+										$cti$ LANGUAGE plpgsql VOLATILE;
+									$$;
+		table_tmpl CONSTANT text := 'CREATE TEMPORARY TABLE %I (LIKE %s)';
```
sumedhpathak: Why does this statement use quotes instead of $$? Could we be consistent?

pg_shard--1.1.sql
```diff
@@ -140,3 +140,51 @@ $sync_table_metadata_to_citus$ LANGUAGE 'plpgsql';
...
+		trg_tmpl CONSTANT text := $$CREATE TRIGGER copy_to_insert
```
sumedhpathak: The func_tmpl has spaces between the $$ and the statement. Could we do something similar here?

Also, what do you think about expanding the names? I especially found trg_tmpl a bit hard to grok.

pg_shard--1.1.sql
jasonmp85: I guess I had a style that I used normal quotes if it fit on one line and $$ for multi-line things. Kind of like how you wouldn’t use a here doc for every string declaration in a shell script…

—Jason

pg_shard--1.1.sql
jasonmp85: It’s so the CREATE FUNCTION falls exactly on a tab stop, which makes tabbing subsequent lines easier.

—Jason

scripts/copy_to_distributed_table.sh
```diff
@@ -0,0 +1,95 @@
...
+options="${options}, FORMAT ${format}"
+filename=$1
+tablename=$2
+
+if [ -z "${filename}" ] || [ -z "${tablename}" ]; then
+	usage
+fi
+
+filename=`echo ${filename} | sed s/\'/\'\'/g`
+facadename=`echo "${tablename}_copy_facade" | sed s/\"/\"\"/g`
```
sumedhpathak: Are we protected against multiple copies into the same table? I noticed in the previous change we also held a lock. 
In this case, the temp table is the same, so running the script twice would error out?

scripts/copy_to_distributed_table.sh
```diff
@@ -0,0 +1,95 @@
...
+/* Don't stop if copy errors out: continue to print file name and row count*/
+\set ON_ERROR_STOP off
+
+\copy pg_temp.${facadename} from ${filename} with ($options)
+
+\o
+
+SELECT currval('rows_inserted');
+
+E_O_SQL
```
sumedhpathak: So at the end of this both the temp table, its trigger and the sequence are all gone/cleaned up? Are there any artifacts left after running this?

scripts/copy_to_distributed_table.sh
jasonmp85: Anything `TEMPORARY` in PostgreSQL is session-bound and session-isolated, i.e. it will disappear at the end of the session and is not visible _at all_ to other sessions. So it is safe to run the script:

  * Multiple times in sequence
  * Simultaneously in multiple processes

The old script used a lock because I didn't know you could make temporary functions. `CREATE TABLE` and `CREATE SEQUENCE` have `TEMPORARY` as part of their syntax, but `CREATE FUNCTION` does not, so I thought that meant no temporary functions. I've since learned that the special `pg_temp` schema always refers to the current session's temporary objects, so creating a function there ensures it cannot collide with anything else and will not live past session end.

In actuality, each session has a numbered temporary schema (e.g. `pg_temp_2`) and `pg_temp` is just an alias for the correct one for your session.

scripts/copy_to_distributed_table.sh
jasonmp85: No artifacts. The UDF creates a temp function, and temp table, and there's a temp sequence involved, so once the session ends everything disappears.

bin/copy_to_distributed_table
```diff
@@ -0,0 +1,163 @@
...
+      Default: the same as the `quote` value (quotes within data are doubled)
+  n : specifies the string that represents a null value
+      Default: \N in text format, an unquoted empty string in CSV format
+  q : specifies the quoting character to be used when a data value is quoted
+      Default: double-quote
+  s : specifies the schema in which the target table resides
+      Default: 'public'
+E_O_USAGE
+
+	exit $1;
```
sumedhpathak: Minor. Should the order of arguments be inverted? (Output device first, followed by exit status?)

bin/copy_to_distributed_table
```diff
@@ -0,0 +1,163 @@
...
+			;;
+		s)
+			schema=`echo ${OPTARG} | sed s/\'/\'\'/g`
+			;;
+		T)
+			format='text'
+			;;
+		*)
+			echo "$0: illegal option -- ${OPTARG}" >&2
+			usage 64 2 # EX_USAGE, STDERR
```
sumedhpathak: Is it worth #defining (or the equivalent in bash) the exit status/output device for readability (apart from the comments?)

sumedhpathak:
I had three higher-level things:

1. I think I am preferring the previous solution of giving a sequence to the function, which it increments. Is there a reason to not do this? i.e., Is there a use-case for the write-through mode in the function apart from being able to install another trigger on the temp-table to count the rows? If not, then I think just passing a sequence makes sense. If no sequence is specified the function doesn't increment it maybe?

I like that the function returns the tablename, so it makes it easier to de-couple the calls and run them manually.

2. Could you add a regression test which supplies a sequence (if you make the above change), and then select's curr_val to validate the function incremented the sequence?

3. Could you also add a regression test after COPY'ing some good rows followed by junk data to validate that the correct rows are recorded as inserted? (Maybe also followed by a select to validate that we inserted the previous 'x' good rows?

bin/copy_to_distributed_table
jasonmp85: I initially just had one argument (status) but realized that commands I consider "well-behaved" output to stdout (and exit 0) when you _ask_ them for help (i.e. pass `-h`) but output to stderr (with non-zero status) when you make some usage mistake. So that necessitated adding a second argument.

I agree that reversing the order makes more sense.

bin/copy_to_distributed_table
jasonmp85: Sure. I need to brush up on scoping rules, but not a bad idea.

BTW, there is no real "standard" for exit codes, except that BSD systems specify a few of them in `sysexit`. I believe the only real standard says "exit between 1 and 126 for error, 0 otherwise", so this is all kind of arbitrary.

At any rate, I like using the BSD ones for things that make sense (usage, missing files, etc.) and make up my own for the rest. At minimum, I consider something well-behaved if:

  * All of the output a user would care about goes to stdout
  * _No_ diagnostic or error output goes to stdout
  * It _always_ exits zero on success (I've seen tools that return counts as their status, which is insane)

Just trying to keep that standard for composability, etc.

jasonmp85:
On each of your high-level points:

  * I agree it is more straightforward to explain, but think it's more modular. It _just_ routes the `INSERT` and knows how to do nothing else. Why pass a sequence? What's special about that, as opposed to—say—passing a function name and having it call that after each row?

  Ultimately, I'm going to change it, but do think the "`INSERT`-only trigger with an option to chain more triggers" is a more composable building block than a "`INSERT` trigger that can update a sequence, if supplied"

  * Yes

  * This is a little trickier. You might have overlooked that I'm not actually _using_ a distributed table in my test. I'm just making a proxy to a normal table. This was because I wanted the test to check the `INSERT` logic alone and not actually set up a `pg_shard` table.

  I find it best to have the behavior that if a feature breaks, _only_ the unit tests related to that feature break. It helps diagnosis. By using a real distributed table, if table distribution breaks, a bunch of tests break, which doesn't aid in finding the problem.

  Of course I'm still playing devil's advocate. I think the `INSERT` trigger is enough of a hack that we _definitely_ want a test to _actually test it against a distributed table_. So I'll update that too, even though it'll complicate the test.

jasonmp85:
Since you didn't mention the naming, I'm assuming you're fine with the names of:

  * The shell script
  * The UDF

I'll make a checklist at the top of this PR to keep track of the things I need to do and when they're done it sounds like this is a :shipit:.

I had one concern: because the UDF returns the proxy table name, I _should_ be able to store it in a variable and interpolate it, or at least I would were I using `COPY` (server-side). From [the documentation](http://www.postgresql.org/docs/9.4/static/app-psql.html) (emphasis mine):

> The syntax of this command is similar to that of the SQL `COPY` command. All options other than the data source/destination are as specified for `COPY`. Because of this, special parsing rules apply to the `\copy` command. _In particular, psql's variable substitution rules and backslash escapes do not apply._

This is why I'm directly interpolating the table name myself in the shell script, which I don't like. Though `\copy` can be slower than `COPY` (data must pass through the client), it has two benefits:

  * It understands relative paths (@rsolari had mentioned this being annoying with the previous solution)
  * You can easily connect from a remote machine and copy over a file _from that machine_ rather than doing it directly on the master
  * Permissions are easier to understand (your user must be able to read the file, not the server)

So `COPY` might be something we explore _if_ the performance of `\copy` is an issue, but otherwise the usability of the latter wins out so I'm sticking with it. I am going to update this to avoid so much direct interpolation, but wanted to call out the tradeoffs in either direction.

sumedhpathak:
@jasonmp85 I am OK with the names of both the script and the UDF.

PR #88, opened 2015-03-14
Modify Citus metadata sync to be idempotent

jasonmp85:
A customer noticed this function could not be called more than once without error. Now calling it more than once has the same effect as calling it once. In addition, if a shard placement's state has changed since the most recent call, an additional call will update the placement state in CitusDB's view.

Added a unit test to verify these cases.

For an explanation of the algorithm in use here, read the _Bulk upsert with lock_ section in [this Stack Overflow answer](http://stackoverflow.com/a/17267423).

**Code review tasks**

  - [x] Fix `JOIN` condition on `UPDATE` to consider node names and ports as well as shard identifier
  - [x] Augment test to place shards on multiple nodes (which would fail without the above change)
  - [x] Clarify comments
  - [x] See whether we can combine any lines, maybe rework formatting

### DIFF
--- a/pg_shard--1.0--1.1.sql
+++ b/pg_shard--1.0--1.1.sql
@@ -6,22 +6,33 @@ LANGUAGE C;
 COMMENT ON FUNCTION partition_column_to_node_string(oid)
 		IS 'return textual form of distributed table''s partition column';
 
-CREATE FUNCTION sync_table_metadata_to_citus(table_name text) RETURNS VOID
+-- Syncs rows from the pg_shard distribution metadata related to the specified
+-- table name into the metadata tables used by CitusDB. After a call to this
+-- function for a particular pg_shard table, that table will become usable for
+-- queries within CitusDB. If placement health has changed for given pg_shard
+-- table, calling this function an additional time will propagate those health
+-- changes to the CitusDB metadata tables.
+CREATE FUNCTION sync_table_metadata_to_citus(table_name text)
+RETURNS void
 AS $sync_table_metadata_to_citus$
 	DECLARE
 		table_relation_id CONSTANT oid NOT NULL := table_name::regclass::oid;
 		dummy_shard_length CONSTANT bigint := 0;
 	BEGIN
-		-- grab lock for upsert
+		-- grab lock to ensure single writer for upsert
 		LOCK TABLE pg_dist_shard_placement IN EXCLUSIVE MODE;
 
-		-- perform update of shard health
+		-- First, update the health of shard placement rows already copied
+		-- from pg_shard to CitusDB. Health is the only mutable attribute,
+		-- so it is presently the only one needing the UPDATE treatment.
 		UPDATE pg_dist_shard_placement
 		SET    shardstate = shard_placement.shard_state
 		FROM   pgs_distribution_metadata.shard_placement
-		WHERE  shardid = shard_placement.shard_id;
+		WHERE  shardid = shard_placement.shard_id AND
+			   nodename = shard_placement.node_name AND
+			   nodeport = shard_placement.node_port;
 
-		-- copy new shard placement metadata
+		-- copy pg_shard placement rows not yet in CitusDB's metadata tables
 		INSERT INTO pg_dist_shard_placement
 					(shardid,
 					 shardstate,
@@ -35,14 +46,15 @@ AS $sync_table_metadata_to_citus$
 			   node_port
 		FROM   pgs_distribution_metadata.shard_placement
 			   LEFT OUTER JOIN pg_dist_shard_placement
-							ON ( shardid = shard_placement.shard_id
-								 AND nodename = shard_placement.node_name
-								 AND nodeport = shard_placement.node_port )
-		WHERE  shardid IS NULL
-			   AND shard_id IN (SELECT id
-								FROM   pgs_distribution_metadata.shard
-								WHERE  relation_id = table_relation_id);
-		-- copy new shard metadata
+							ON ( shardid = shard_placement.shard_id AND
+								 nodename = shard_placement.node_name AND
+								 nodeport = shard_placement.node_port )
+		WHERE  shardid IS NULL AND
+			   shard_id IN (SELECT id
+							FROM   pgs_distribution_metadata.shard
+							WHERE  relation_id = table_relation_id);
+
+		-- copy pg_shard shard rows not yet in CitusDB's metadata tables
 		INSERT INTO pg_dist_shard
 					(shardid,
 					 logicalrelid,
@@ -57,11 +69,13 @@ AS $sync_table_metadata_to_citus$
 		FROM   pgs_distribution_metadata.shard
 			   LEFT OUTER JOIN pg_dist_shard
 							ON ( shardid = shard.id )
-		WHERE  shardid IS NULL
-			   AND relation_id = table_relation_id;
+		WHERE  shardid IS NULL AND
+			   relation_id = table_relation_id;
 
-		-- copy new partition metadata, which also converts the partition column to
-		-- a node string representation as expected by CitusDB
+		-- Finally, copy pg_shard partition rows not yet in CitusDB's metadata
+		-- tables. CitusDB uses a textual form of a Var node representing the
+		-- partition column, so we must use a special function to transform the
+		-- representation used by pg_shard (which is just the column name).
 		INSERT INTO pg_dist_partition
 					(logicalrelid,
 					 partmethod,
@@ -72,8 +86,8 @@ AS $sync_table_metadata_to_citus$
 		FROM   pgs_distribution_metadata.partition
 			   LEFT OUTER JOIN pg_dist_partition
 							ON ( logicalrelid = partition.relation_id )
-		WHERE  logicalrelid IS NULL
-			   AND relation_id = table_relation_id;
+		WHERE  logicalrelid IS NULL AND
+			   relation_id = table_relation_id;
 	END;
 $sync_table_metadata_to_citus$ LANGUAGE 'plpgsql';
 
@@ -123,7 +137,7 @@ AS $create_insert_proxy_for_table$
 		INTO   STRICT attr_names
 		FROM   pg_attribute
 		WHERE  attrelid = target_table AND
-			   attnum > 0          AND
+			   attnum > 0 AND
 			   NOT attisdropped;
 
 		-- build fully specified column list and USING clause from attr. names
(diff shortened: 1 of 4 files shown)


pg_shard--1.0--1.1.sql
```diff
@@ -12,7 +12,16 @@ AS $sync_table_metadata_to_citus$
 		table_relation_id CONSTANT oid NOT NULL := table_name::regclass::oid;
 		dummy_shard_length CONSTANT bigint := 0;
 	BEGIN
-		-- copy shard placement metadata
+		-- grab lock for upsert
+		LOCK TABLE pg_dist_shard_placement IN EXCLUSIVE MODE;
+
+		-- perform update of shard health
+		UPDATE pg_dist_shard_placement
```
onderkalaci: I think we need to consider two things here:
* Shard placements must be identified not by shardid alone. We should use (shard_id + nodename + nodeport)
* Also, we need to consider only the shards that belong to table_relation_id.

So, I think this statement should look like as below:
```sql
-- perform update of shard health
UPDATE pg_dist_shard_placement
SET    shardstate = shard_placement.shard_state
FROM   pgs_distribution_metadata.shard_placement
WHERE  shardid = shard_placement.shard_id AND
       nodename = shard_placement.node_name AND
       nodeport = shard_placement.node_port AND
       shardid IN (SELECT id FROM   pgs_distribution_metadata.shard
		         WHERE  relation_id = table_relation_id);
```

pg_shard--1.0--1.1.sql
```diff
@@ -12,7 +12,16 @@ AS $sync_table_metadata_to_citus$
...
+		SET    shardstate = shard_placement.shard_state
+		FROM   pgs_distribution_metadata.shard_placement
+		WHERE  shardid = shard_placement.shard_id;
+
+		-- copy new shard placement metadata
 		INSERT INTO pg_dist_shard_placement
```
onderkalaci: I think what  @ozgune suggests as follows:
```sql
INSERT INTO pg_dist_shard_placement (shardid, shardstate, shardlength,nodename, nodeport
```

pg_shard--1.0--1.1.sql
```diff
@@ -25,11 +34,15 @@ AS $sync_table_metadata_to_citus$
...
+			   LEFT OUTER JOIN pg_dist_shard_placement
+							ON ( shardid = shard_placement.shard_id
+								 AND nodename = shard_placement.node_name
+								 AND nodeport = shard_placement.node_port )
+		WHERE  shardid IS NULL
+			   AND shard_id IN (SELECT id
+								FROM   pgs_distribution_metadata.shard
+								WHERE  relation_id = table_relation_id);
+		-- copy new shard metadata
 		INSERT INTO pg_dist_shard
```
onderkalaci: Again we can consider ozgun's suggestion.

pg_shard--1.0--1.1.sql
```diff
@@ -42,9 +55,12 @@ AS $sync_table_metadata_to_citus$
...
-		WHERE  relation_id = table_relation_id;
+			   LEFT OUTER JOIN pg_dist_shard
+							ON ( shardid = shard.id )
+		WHERE  shardid IS NULL
+			   AND relation_id = table_relation_id;
 
-		-- copy partition metadata, which also converts the partition column to
+		-- copy new partition metadata, which also converts the partition column to
 		-- a node string representation as expected by CitusDB
 		INSERT INTO pg_dist_partition
```
onderkalaci: Again we can consider ozgun's suggestion.

pg_shard--1.0--1.1.sql
```diff
@@ -42,9 +55,12 @@ AS $sync_table_metadata_to_citus$
 			   min_value,
```
onderkalaci: Again we can consider ozgun's suggestion.

pg_shard--1.1.sql
```diff
@@ -92,7 +92,16 @@ AS $sync_table_metadata_to_citus$
...
+		LOCK TABLE pg_dist_shard_placement IN EXCLUSIVE MODE;
+
+		-- perform update of shard health
+		UPDATE pg_dist_shard_placement
+		SET    shardstate = shard_placement.shard_state
+		FROM   pgs_distribution_metadata.shard_placement
+		WHERE  shardid = shard_placement.shard_id;
+
+		-- copy new shard placement metadata
 		INSERT INTO pg_dist_shard_placement
```
onderkalaci: Again we can consider ozgun's suggestion.

pg_shard--1.1.sql
```diff
@@ -105,11 +114,15 @@ AS $sync_table_metadata_to_citus$
...
+			   LEFT OUTER JOIN pg_dist_shard_placement
+							ON ( shardid = shard_placement.shard_id
+								 AND nodename = shard_placement.node_name
+								 AND nodeport = shard_placement.node_port )
+		WHERE  shardid IS NULL
+			   AND shard_id IN (SELECT id
+								FROM   pgs_distribution_metadata.shard
+								WHERE  relation_id = table_relation_id);
+		-- copy new shard metadata
 		INSERT INTO pg_dist_shard
```
onderkalaci: Again we can consider ozgun's suggestion.

pg_shard--1.1.sql
```diff
@@ -122,9 +135,12 @@ AS $sync_table_metadata_to_citus$
...
-		WHERE  relation_id = table_relation_id;
+			   LEFT OUTER JOIN pg_dist_shard
+							ON ( shardid = shard.id )
+		WHERE  shardid IS NULL
+			   AND relation_id = table_relation_id;
 
-		-- copy partition metadata, which also converts the partition column to
+		-- copy new partition metadata, which also converts the partition column to
 		-- a node string representation as expected by CitusDB
 		INSERT INTO pg_dist_partition
```
onderkalaci: Again we can consider ozgun's suggestion.

pg_shard--1.0--1.1.sql
```diff
@@ -12,7 +12,16 @@ AS $sync_table_metadata_to_citus$
 		table_relation_id CONSTANT oid NOT NULL := table_name::regclass::oid;
 		dummy_shard_length CONSTANT bigint := 0;
 	BEGIN
-		-- copy shard placement metadata
+		-- grab lock for upsert
+		LOCK TABLE pg_dist_shard_placement IN EXCLUSIVE MODE;
+
+		-- perform update of shard health
```
onderkalaci: As far as I understand, we aim to 
* Firstly update existing shard placements in pg_dist_shard_placement
* Then, add the shard_placements that does not exist in pg_dist_shard_placement,  pg_dist_shard and  pg_dist_partition.

So, updating the comments that explains this could be better, what do you think?
1) -- perform update of shard health => -- perform update of shard health which are already synced
2) -- copy new shard placement metadata => -- copy shard placement metadata which does not exist in pg_dist_shard_placement
3) -- copy new shard metadata  =>  copy shard metadata which does not exist in pg_dist_shard
4) -- copy new partition metadata, which also converts the partition column to ...   => copy partition metadata which does not exist in pg_dist_partition ..

pg_shard--1.1.sql
```diff
@@ -92,7 +92,16 @@ AS $sync_table_metadata_to_citus$
 		table_relation_id CONSTANT oid NOT NULL := table_name::regclass::oid;
 		dummy_shard_length CONSTANT bigint := 0;
 	BEGIN
-		-- copy shard placement metadata
+		-- grab lock for upsert
```
onderkalaci: I only commented to  pg_shard--1.0--1.1.sql, all of which are also valid here.

onderkalaci:
@jasonmp85 For style issues of SQL I am not much experienced. Also, we don't have a documentation for SQL styling.
I only added commented for the style that Ozgun already suggested.
In general, I have a comment. Compared to CitusDB SQL examples, we should maybe not start a new line on each AND clause in the WHERE statements. For example, instead of
```sql
		WHERE  shardid IS NULL
			   AND relation_id = table_relation_id;
```
we may consider the following to make it similar to CitusDB:
```sql
		WHERE  shardid IS NULL  AND relation_id = table_relation_id;
```
But, I am not sure about that, so I didn't comment on this cases. Just write here to discuss.

pg_shard--1.0--1.1.sql
```diff
@@ -12,7 +12,16 @@ AS $sync_table_metadata_to_citus$
 		table_relation_id CONSTANT oid NOT NULL := table_name::regclass::oid;
 		dummy_shard_length CONSTANT bigint := 0;
 	BEGIN
-		-- copy shard placement metadata
+		-- grab lock for upsert
+		LOCK TABLE pg_dist_shard_placement IN EXCLUSIVE MODE;
```
onderkalaci: Why not simple use ACCESS EXCLUSIVE and disallow reads too? Would it be too restrictive?

pg_shard--1.0--1.1.sql
jasonmp85: Good catch! I had copy-pasted the `JOIN` clause onto most of the statements but eventually caught that the placements should have an equality clause on the triple you mention, not just the shard identifier.

I fixed it in the `INSERT` but you're right that I missed the `UPDATE`. Fixing (and I'll add a unit test that'll fail without this).

pg_shard--1.0--1.1.sql
jasonmp85: We use 90 characters as the limit on everything except git commit messages… I think we also use tabs (four columns) everywhere to be consistent. We don't really have an SQL style specified, so I used [this little web page](http://www.dpriver.com/pp/sqlformat.htm) to format things in v1.0 just so they'd have some sort of consistency without us really having to think much about it.

I'll look into whether there are places I can join lines to make them longer, but I want to stick with the output of this tool for now, since previous checkins used its style. That way we can keep line noise/churn down.

pg_shard--1.0--1.1.sql
jasonmp85: The [StackOverflow answer](http://stackoverflow.com/a/17267423) I was referencing used `EXCLUSIVE MODE` and I tend to agree with it. If the CitusDB metadata tables are out of sync with the `pg_shard` ones, then any queries processed by CitusDB will potentially hit stale shards, so one could make the argument that we should lock them out until the update is complete.

But because right now this sync function requires manual intervention anyways, they will _always_ read stale metadata until a human runs this function, so I think it's fine to let them go ahead and keep reading metadata while we are updating it. We just need to ensure the upsert itself is serialized, which is what the lock provides.

pg_shard--1.0--1.1.sql
jasonmp85: I made minimal changes to the comments, but you're right, they can be more clear. I'll update them

jasonmp85:
Since I want to get this wrapped up, I'm assuming after these points are addressed, this is a :shipit:… if not, can duly gripe at me and I'll fix any other issues in another branch :grimacing:.

PR #89, opened 2015-03-14
Add uncrustify config and use it

jasonmp85:
Works well with at least Uncrustify 0.61, though it may work just as well with other versions. In zsh one can check style with:

    uncrustify -c .uncrustify.cfg --check  **/*.[ch]~ruleutils*

This'll exit with 0 if all files are compliant and will exit with 1 otherwise. Pass --no-backup instead of --check to have uncrustify modify all files in place to be compliant.

Deduced style from most of our codebase and style guides and consulted with team members for any ambiguities I needed resolved.

### DIFF
--- a/.uncrustify.cfg
+++ b/.uncrustify.cfg
@@ -1,0 +1,1693 @@
+# Uncrustify 0.61
+
+#
+# General options
+#
+
+# The type of line endings
+newlines                                  = lf     # auto/lf/crlf/cr
+
+# The original size of tabs in the input
+input_tab_size                            = 4        # number
+
+# The size of tabs in the output (only used if align_with_tabs=true)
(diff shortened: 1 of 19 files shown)


jasonmp85:
To _just_ see the non-whitespace changes to the code made by this branch, check out [this link](https://github.com/citusdata/pg_shard/commit/68dffa363d6bc83140cf23679381c2d0012ff41b?ts=4&w=1). I may add `uncrustify --check` to our Travis build in the near future to reject changes that are not compliant with the style rules. They're pretty lenient, and will give you leeway to add breaks, etc. where you see fit, but will make sure you don't have trailing space, mix tabs, misalign variables, etc. We can edit them as we go forward.

PR #92, opened 2015-03-20
Support composite types / enumeration types / varchar for partition columns

### LINKED ISSUE #80: Support enumeration types for partition columns
This is similar to #79 except maybe more surprising that it doesn't _Just Work™_:

```sql
CREATE TYPE bug_status AS ENUM ('new', 'open', 'closed');
# CREATE TYPE
# Time: 6.044 ms

CREATE TABLE bugs (
    id integer,
    status bug_status
);
# CREATE TABLE
# Time: 3.599 ms

SELECT master_create_distributed_table('bugs', 'status');
# ERROR:  42883: could not identify a hash function for type bug_status
# DETAIL:  Partition column types must have a hash function defined to use hash partitioning.
# LOCATION:  master_create_distributed_table, create_shards.c:109
# Time: 3.202 ms
```

Kinda weird that enum types don't have a built-in hash function defined, given they're basically glorified integers.

onderkalaci:
Hey @jasonmp85 ,

This PR aims to handle how we find the operatorId for the partition columns. Previously, we find operators by directly using the partition column's type.
However, it was wrong. We need to find operators by using the partition column's opfamily input type. The reason is that we use "get_opfamily_member()" to get the operators. As it name implies,
that function finds the operators for opfamily. So, providing the opclass' input type makes more sense. 

With this PR, I tested with different partition columns (distribute table, test INSERT/SELECT/UPDATE):

Test No | Test Type | STATUS
------------- | -------------  | ------------- 
1) | Test with varchar														| PASSED
2) |Test with enum types														|PASSED
3) |Test with basic types (text)												|PASSED
4) |Test with arrays of basic type											|PASSED
5) |Test with arrays of enum													|PASSED
6) |Test with composite types with hash function defined						|PASSED
7) |Test with arrays of composite types with hash function defined			|PASSED
8) |Test with json type														|PASSED (cannot distribute table since it has no default op class)
9) |Test with jsonb type														|PASSED

This fix is based on the changes on CitusDB (Composite type bug). However, there is a difference on the implementation compared to that fix. On CitusDB, we changed "*MakeOpExpression*" function too.
However, that is not needed for us since we already use hashed(integers) values on "*MakeOpExpression*". But, when we support range partitioning, we may need to add that change too. I thought adding it now, but that would seem very unrelated.

Fixes #80


onderkalaci:
@jasonmp85 This is not ready to be merged. After applying the changes in the [pull request](https://github.com/citusdata/pg_shard/pull/102), I observer a different problem.

We are successfully executing UPDATE/DELETE on the composite type partitioned table such as:

```sql
UPDATE comp_distributed_table SET id=12  where col = '(1,2)';
```
However, during INSERTs we have the following error
```sql
INSERT INTO comp_distributed_table VALUES(1, '(5,6)');
ERROR:  cannot modify multiple shards during a single query
```
So, lets not merge this yet. I'll need to have a look on this problem more.

onderkalaci:
Hey @jasonmp85 ,

I added a new file for tests. My motivation was to that we have to both create shards and execute queries on that shard. So, putting these tests to create_shards.sql or queries.sql separately didn't fit as good as putting them in a separate file. 

What do you think? Is it OK ?

jasonmp85:
This looks great. Cleaning up and merging.

PR #102, opened 2015-04-13
Bugfix volalite functions#47

### LINKED ISSUE #47: VOLATILE functions accepted in non-partition key qual. clauses

onderkalaci:
This pull request prevents WHERE clause statements to have VOLATILE or MUTABLE functions.

Fixes #47

**Review tasks**

  - [x] Fix comment in test
  - [x] Rework area that used ternary expression
  - [x] Remove call to `contain_volatile_functions`
  - [x] Add test with mutable but not volatile function such as `current_timestamp`

### DIFF
--- a/expected/modifications.out.tmpl
+++ b/expected/modifications.out.tmpl
@@ -77,6 +77,9 @@ ERROR:  cannot plan sharded modification containing values which are not constan
 -- commands with expressions that cannot be collapsed are unsupported
 INSERT INTO limit_orders VALUES (2036, 'GOOG', 5634, now(), 'buy', random());
 ERROR:  cannot plan sharded modification containing values which are not constants or constant expressions
+-- commands with mutable or volitale functions 
+DELETE FROM limit_orders WHERE id = 246 AND bidder_id = (random() * 1000);
+ERROR:  cannot plan sharded modification containing values which are not constants or constant expressions
 -- commands with multiple rows are unsupported
 INSERT INTO limit_orders VALUES (DEFAULT), (DEFAULT);
 ERROR:  cannot perform distributed planning for the given query
--- a/expected/modifications.out.tmpl
+++ b/expected/modifications.out.tmpl
@@ -80,7 +80,7 @@ ERROR:  cannot plan sharded modification containing values which are not constan
 -- commands with mutable functions in their quals
 DELETE FROM limit_orders WHERE id = 246 AND bidder_id = (random() * 1000);
 ERROR:  cannot plan sharded modification containing values which are not constants or constant expressions
--- commands with mutable but non-volatilte functions in their quals
+-- commands with mutable but non-volatilte functions(ie: stable func.) in their quals
 DELETE FROM limit_orders WHERE id = 246 AND placed_at = current_timestamp;
 ERROR:  cannot plan sharded modification containing values which are not constants or constant expressions
 -- commands with multiple rows are unsupported
--- a/expected/modifications.out.tmpl
+++ b/expected/modifications.out.tmpl
@@ -77,8 +77,11 @@ ERROR:  cannot plan sharded modification containing values which are not constan
 -- commands with expressions that cannot be collapsed are unsupported
 INSERT INTO limit_orders VALUES (2036, 'GOOG', 5634, now(), 'buy', random());
 ERROR:  cannot plan sharded modification containing values which are not constants or constant expressions
--- commands with mutable or volitale functions 
+-- commands with mutable functions in their quals
 DELETE FROM limit_orders WHERE id = 246 AND bidder_id = (random() * 1000);
+ERROR:  cannot plan sharded modification containing values which are not constants or constant expressions
+-- commands with mutable but non-volatilte functions in their quals
+DELETE FROM limit_orders WHERE id = 246 AND placed_at = current_timestamp;
 ERROR:  cannot plan sharded modification containing values which are not constants or constant expressions
 -- commands with multiple rows are unsupported
 INSERT INTO limit_orders VALUES (DEFAULT), (DEFAULT);
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -393,7 +393,9 @@ ErrorIfQueryNotSupported(Query *queryTree)
 	bool hasValuesScan = false;
 	uint32 queryTableCount = 0;
 	bool hasNonConstTargetEntryExprs = false;
+	bool hasNonConstQualExprs = false;
 	bool specifiesPartitionValue = false;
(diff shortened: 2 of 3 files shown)


onderkalaci:
@jasonmp85 I think we can do a similar change for [this line](https://github.com/citusdata/pg_shard/blob/develop/pg_shard.c#L530). In other words, instead of checking 
```c
!IsA(targetEntry->expr, Const)
```
We can do something like:
```c
contain_mutable_functions((targetEntry->exp) ||
	contain_volatile_functions((targetEntry->exp))	
```
But, since the issue only considers QUALs, I didn't want to do that here.

pg_shard.c
```diff
@@ -393,7 +393,9 @@ ErrorIfQueryNotSupported(Query *queryTree)
 	bool hasValuesScan = false;
 	uint32 queryTableCount = 0;
 	bool hasNonConstTargetEntryExprs = false;
+	bool hasNonConstQualExprs = false;
 	bool specifiesPartitionValue = false;
+	Node *whereClause = NULL;
```
jasonmp85: This variable can have its scope reduced… please put it inside the `if (commandType == CMD_INSERT || commandType == CMD_UPDATE ||…` block.

pg_shard.c
```diff
@@ -537,9 +539,17 @@ ErrorIfQueryNotSupported(Query *queryTree)
 				specifiesPartitionValue = true;
 			}
 		}
+
+		whereClause = ((queryTree->jointree == NULL) ?
```
jasonmp85: We _seldom_ (never, maybe?) use the ternary operator in our style. I'd say this is more idiomatic CitusDB style:

```c
FromExpr *joinTree = query->jointree;
if (joinTree != NULL && contain_volatile_functions(joinTree->quals))
{
    hasNonConstQualExprs = true;
}
```

pg_shard.c
```diff
@@ -537,9 +539,17 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
+					   NULL : queryTree->jointree->quals);
+		if (contain_mutable_functions(whereClause) ||
+			contain_volatile_functions(whereClause))
```
jasonmp85: I think it's sufficient to call `contain_mutable_function` here: anything that is `VOLATILE` is also not `IMMUTABLE`, so this is redundant.

expected/modifications.out.tmpl
```diff
@@ -77,6 +77,9 @@ ERROR:  cannot plan sharded modification containing values which are not constan
 -- commands with expressions that cannot be collapsed are unsupported
 INSERT INTO limit_orders VALUES (2036, 'GOOG', 5634, now(), 'buy', random());
 ERROR:  cannot plan sharded modification containing values which are not constants or constant expressions
+-- commands with mutable or volitale functions
```
jasonmp85: Misspelled, but replace with `commands with mutable functions in their quals`

jasonmp85:
@onderkalaci this looks good and after the small style changes (and removing the redundant call), it's a :shipit:.

jasonmp85:
@onderkalaci on the topic of

jasonmp85:
Accidentally closed, sorry!

jasonmp85:
@onderkalaci on the topic of `!IsA(targetEntry->expr, Const)`, we already know any function calls we see in the target list _must_ be mutable. This is because we have already called the `standard_planner` on the query. That in turn calls `eval_const_expressions`, which eagerly replaces any immutable functions with their output (this also applies to the quals; for instance, doing an `UPDATE … WHERE foo = (1 + 1)` becomes `UPDATE … WHERE foo = 2`).

So for target lists, we only accept constants, but for quals, there might be a column reference (LHS), operator, and value (RHS). In that case, the standard planner may have collapsed the RHS to a constant (if, as above it's something like `1+1`), but the operator call is still there. But e.g. `=` is `IMMUTABLE` so it needs to make it through.

jasonmp85:
So I guess I'm saying the walker is overkill for the target list but is needed for the quals, so what we have here is now complete.

pg_shard.c
jasonmp85: Instead of `whereClause`, I think you just need to declare `FromExpr *joinTree;` inside the `if` block.

sql/modifications.sql
```diff
@@ -54,6 +54,9 @@ INSERT INTO limit_orders VALUES (random() * 100, 'ORCL', 152, '2011-08-25 11:50:
 -- commands with expressions that cannot be collapsed are unsupported
 INSERT INTO limit_orders VALUES (2036, 'GOOG', 5634, now(), 'buy', random());
 
+-- commands with mutable or volitale functions 
+DELETE FROM limit_orders WHERE id = 246 AND bidder_id = (random() * 1000);
+
```
jasonmp85: Can you add another test with a function which is stable but not mutable? That way we'll have one test that checks `VOLATILE` and another checking `STABLE`.

jasonmp85:
Added checklist up to. Fix and submit the changes and I'll merge tomorrow if all looks good.

onderkalaci:
@jasonmp85 all your comments make sense and I applied the changes.

jasonmp85:
Looks good. Merging.

PR #103, opened 2015-04-15
CitusDB metadata interoperability

### LINKED ISSUE #11: Citus treats pg_shard tables as local when missing metadata
_From @sumedhpathak on November 20, 2014 0:27_

If someone doesn't copy over the metadata from pg_shard to CitusDB, but sets `use_citusdb_select_logic` to `true`, in CitusDB we treat all pg_shard tables as local tables and read the empty table on the master node, returning 0 results.

This is minor, but can be a usability issue with CitusDB/pg_shard integration.

### LINKED ISSUE #27: Plan bidirectional metadata sync
Citus and pg_shard should be able to sync distribution metadata back and forth without any user intervention. Determine what approaches could meet this need.

jasonmp85:
OK, so this has been entirely rebased on top of the changes from #110, i.e. we're using SPI everywhere now. This review is now all about the triggers and views needed to adapt CitusDB's metadata tables to be used within `pg_shard`.

Travis is building this under PostgreSQL 9.3/9.4, and CitusDB 4.0, so there is some assurance this actually works.

**Code Review Tasks**

  - [x] Add initialization mode to assign variables based on CitusDB presence
  - [x] Figure out how to test this
  - [x] Transform `pg_shard` tables into views when Citus present (to prevent them from appearing blank), or just error out when queried.
  - [x] Determine ramifications for upgrading from `pg_shard` to CitusDB. Users might `pg_dump` from a PostgreSQL `pg_shard` install and try to load into CitusDB. This must work seamlessly.
  - [x] Determine interoperability requirements for `\STAGE` and using an existing CitusDB-distributed table with `pg_shard`
  - [x] Address interoperability issues

Resolves #11.
Resolves #27.

### DIFF
--- a/distribution_metadata.c
+++ b/distribution_metadata.c
@@ -23,6 +23,7 @@
 #include "access/htup.h"
 #include "access/tupdesc.h"
 #include "executor/spi.h"
+#include "catalog/catalog.h"
 #include "catalog/namespace.h"
 #include "catalog/pg_type.h"
 #include "nodes/makefuncs.h"
@@ -374,6 +375,7 @@ PartitionType(Oid distributedTableId)
 bool
 IsDistributedTable(Oid tableId)
 {
+	Oid tableNamespaceOid = get_rel_namespace(tableId);
 	Oid metadataNamespaceOid = get_namespace_oid("pgs_distribution_metadata", false);
 	Oid partitionMetadataTableOid = get_relname_relid("partition", metadataNamespaceOid);
 	bool isDistributedTable = false;
@@ -386,7 +388,7 @@ IsDistributedTable(Oid tableId)
 	 * The query below hits the partition metadata table, so if we don't detect
 	 * that and short-circuit, we'll get infinite recursion in the planner.
 	 */
-	if (tableId == partitionMetadataTableOid)
+	if (IsSystemNamespace(tableNamespaceOid) || tableId == partitionMetadataTableOid)
 	{
 		return false;
 	}
--- a/distribution_metadata.c
+++ b/distribution_metadata.c
@@ -521,19 +521,33 @@ TupleToShardInterval(HeapTuple heapTuple, TupleDesc tupleDescriptor)
 	Oid relationId = DatumGetObjectId(relationIdDatum);
 
 	char partitionType = DatumGetChar(partitionTypeDatum);
-	if (partitionType == HASH_PARTITION_TYPE)
+	switch (partitionType)
 	{
-		intervalTypeId = INT4OID;
-	}
-	else
-	{
-		Datum keyDatum = SPI_getbinval(heapTuple, tupleDescriptor,
-									   TLIST_NUM_SHARD_KEY, &isNull);
-		char *partitionColumnName = TextDatumGetCString(keyDatum);
+		case APPEND_PARTITION_TYPE:
+		case RANGE_PARTITION_TYPE:
+		{
+			Datum keyDatum = SPI_getbinval(heapTuple, tupleDescriptor,
+										   TLIST_NUM_SHARD_KEY, &isNull);
+			char *partitionColumnName = TextDatumGetCString(keyDatum);
 
-		Var *partitionColumn = ColumnNameToColumn(relationId, partitionColumnName);
-		intervalTypeId = partitionColumn->vartype;
-		intervalTypeMod = partitionColumn->vartypmod;
+			Var *partitionColumn = ColumnNameToColumn(relationId, partitionColumnName);
+			intervalTypeId = partitionColumn->vartype;
+			intervalTypeMod = partitionColumn->vartypmod;
+			break;
+		}
+
+		case HASH_PARTITION_TYPE:
+		{
+			intervalTypeId = INT4OID;
+			break;
+		}
+
+		default:
+		{
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("unsupported table partition type: %c",
+								   partitionType)));
+		}
 	}
 
 	getTypeInputInfo(intervalTypeId, &inputFunctionId, &typeIoParam);
--- a/distribution_metadata.c
+++ b/distribution_metadata.c
@@ -370,24 +370,36 @@ PartitionType(Oid distributedTableId)
 
 
 /*
- * IsDistributedTable simply returns whether the specified table is distributed.
+ * IsDistributedTable returns whether the specified table is distributed. It
+ * returns false if the input is InvalidOid.
  */
 bool
 IsDistributedTable(Oid tableId)
 {
-	Oid tableNamespaceOid = get_rel_namespace(tableId);
 	Oid metadataNamespaceOid = get_namespace_oid("pgs_distribution_metadata", false);
-	Oid partitionMetadataTableOid = get_relname_relid("partition", metadataNamespaceOid);
+	Oid tableNamespaceOid = InvalidOid;
+	Oid partitionMetadataTableOid = InvalidOid;
 	bool isDistributedTable = false;
 	Oid argTypes[] = { OIDOID };
 	Datum argValues[] = { ObjectIdGetDatum(tableId) };
 	const int argCount = sizeof(argValues) / sizeof(argValues[0]);
 	int spiStatus PG_USED_FOR_ASSERTS_ONLY = 0;
 
+	/* short-circuit if the input is invalid */
+	if (tableId == InvalidOid)
+	{
+		return false;
+	}
+
 	/*
 	 * The query below hits the partition metadata table, so if we don't detect
 	 * that and short-circuit, we'll get infinite recursion in the planner.
+	 *
+	 * Within CitusDB, a view rewrite the query to reference CitusDB catalogs,
+	 * so we also need to catch whether the table exists in a system namespace.
 	 */
+	tableNamespaceOid = get_rel_namespace(tableId);
+	partitionMetadataTableOid = get_relname_relid("partition", metadataNamespaceOid);
 	if (IsSystemNamespace(tableNamespaceOid) || tableId == partitionMetadataTableOid)
 	{
 		return false;
--- a/pg_shard--1.1--1.2.sql
+++ b/pg_shard--1.1--1.2.sql
@@ -1,13 +1,242 @@
--- add default values to id columns
-ALTER TABLE pgs_distribution_metadata.shard
-	ALTER COLUMN id
-	SET DEFAULT nextval('pgs_distribution_metadata.shard_id_sequence');
-ALTER TABLE pgs_distribution_metadata.shard_placement
-	ALTER COLUMN id
-	SET DEFAULT nextval('pgs_distribution_metadata.shard_placement_id_sequence');
+-- needed in our views
+CREATE FUNCTION column_to_column_name(table_oid oid, column_var text)
+RETURNS text
+AS 'MODULE_PATHNAME'
+LANGUAGE C STABLE STRICT;
 
--- associate sequences with their columns
-ALTER SEQUENCE pgs_distribution_metadata.shard_id_sequence
-	OWNED BY pgs_distribution_metadata.shard.id;
-ALTER SEQUENCE pgs_distribution_metadata.shard_placement_id_sequence
-	OWNED BY pgs_distribution_metadata.shard_placement.id;
+CREATE FUNCTION column_name_to_column(table_oid oid, column_name text)
+RETURNS text
+AS 'MODULE_PATHNAME'
+LANGUAGE C STABLE STRICT;
+
+DO $$
+DECLARE
+	use_citus_metadata boolean := false;
+	relation_name oid;
+BEGIN
+	BEGIN
+		PERFORM 'pg_catalog.pg_dist_partition'::regclass;
+		use_citus_metadata = true;
+	EXCEPTION
+		WHEN undefined_table THEN
+			use_citus_metadata = false;
+	END;
+
+	IF use_citus_metadata THEN
+		-- just in case, lock everyone out of pg_shard partitions
+		LOCK TABLE pgs_distribution_metadata.partition IN EXCLUSIVE MODE;
+		FOR relation_name IN SELECT relation_id::regclass::text
+		FROM pgs_distribution_metadata.partition LOOP
+			PERFORM sync_table_metadata_to_citus(relation_id::regclass);
+		END LOOP;
+
+		DROP TABLE pgs_distribution_metadata.partition,
+				   pgs_distribution_metadata.shard,
+				   pgs_distribution_metadata.shard_placement;
+
+		CREATE FUNCTION adapt_and_insert_shard() RETURNS TRIGGER AS $aais$
+		BEGIN
+			IF NEW.id IS NULL THEN
+				NEW.id = nextval('pg_dist_shardid_seq');
+			END IF;
+
+			INSERT INTO pg_dist_shard
+						(logicalrelid,
+						 shardid,
+						 shardstorage,
+						 shardalias,
+						 shardminvalue,
+						 shardmaxvalue)
+			VALUES      (NEW.relation_id,
+						 NEW.id,
+						 NEW.storage,
+						 NULL,
+						 NEW.min_value,
+						 NEW.max_value);
+
+			RETURN NEW;
+		END
+		$aais$ LANGUAGE plpgsql;
+
+		CREATE FUNCTION adapt_and_insert_shard_placement() RETURNS trigger AS $aaisp$
+		BEGIN
+			INSERT INTO pg_dist_shard_placement
+						(shardid,
+						 shardstate,
+						 shardlength,
+						 nodename,
+						 nodeport)
+			VALUES      (NEW.shard_id,
+						 NEW.shard_state,
+						 0,
+						 NEW.node_name,
+						 NEW.node_port)
+			RETURNING oid INTO STRICT NEW.id;
+
+			RETURN NEW;
+		END
+		$aaisp$ LANGUAGE plpgsql;
+
+		CREATE FUNCTION adapt_and_insert_partition() RETURNS trigger AS $aaip$
+		BEGIN
+			INSERT INTO pg_dist_partition
+						(logicalrelid,
+						 partmethod,
+						 partkey)
+			VALUES      (NEW.relation_id,
+						 NEW.partition_method,
+						 column_name_to_column(NEW.relation_id, NEW.key));
+
+			RETURN NEW;
+		END
+		$aaip$ LANGUAGE plpgsql;
+
+		-- metadata relations are views under CitusDB
+		CREATE VIEW pgs_distribution_metadata.shard AS
+			SELECT shardid       AS id,
+				   logicalrelid  AS relation_id,
+				   shardstorage  AS storage,
+				   shardminvalue AS min_value,
+				   shardmaxvalue AS max_value
+			FROM   pg_dist_shard;
+
+		CREATE TRIGGER shard_insert
+		INSTEAD OF INSERT ON pgs_distribution_metadata.shard
+			FOR EACH ROW
+			EXECUTE PROCEDURE adapt_and_insert_shard();
+
+		CREATE VIEW pgs_distribution_metadata.shard_placement AS
+			SELECT oid::bigint AS id,
+				   shardid     AS shard_id,
+				   shardstate  AS shard_state,
+				   nodename    AS node_name,
+				   nodeport    AS node_port
+			FROM   pg_dist_shard_placement;
+
+		CREATE TRIGGER shard_placement_insert
+		INSTEAD OF INSERT ON pgs_distribution_metadata.shard_placement
+			FOR EACH ROW
+			EXECUTE PROCEDURE adapt_and_insert_shard_placement();
+
+		CREATE VIEW pgs_distribution_metadata.partition AS
+			SELECT logicalrelid AS relation_id,
+				   partmethod   AS partition_method,
+				   column_to_column_name(logicalrelid, partkey) AS key
+			FROM   pg_dist_partition;
+
+		CREATE TRIGGER partition_insert
+		INSTEAD OF INSERT ON pgs_distribution_metadata.partition
+			FOR EACH ROW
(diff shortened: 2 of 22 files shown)


marcocitus:
I'm aware that this work in progress, but some early feedback. We were trying this out today to see if it would help make the CitusDB -> pg_shard migration path easier for CitusDB documentation purposes, since that seems to be the more common direction. We created a table using CitusDB with DISTRIBUTE BY APPEND and used this branch of pg_shard.

One problem we ran into is that when doing an INSERT in an non-existing range we got "ERROR:  cannot execute INSERT on a distributed table on master node", because pg_shard ExecutorStart executor hook falls back to CitusDB for zero-shard queries. This led to some confusion.

The next problem we ran into is that pg_shard caches the metadata. Any change made to the metadata by CitusDB is not visible to pg_shard until a new session is started, this also led to some confusion. For example, performing \STAGE and then INSERT into the range would give "ERROR:  no placements exist for shard with ID". 

Finally, we added a new shard using \STAGE, the range of which overlapped with an existing the shard. We then tried an INSERT on the overlapping range. It went to the first shard. We then started a new session to clear the cache, after which we got "ERROR:  cannot modify multiple shards during a single query". This was expected, but I guess an INSERT should always go to at most one shard.

It seems the more fundamental issue here is the caching. We might have to make changes to CitusDB to help pg_shard clear its cache or set up a trigger on the catalog table.

jasonmp85:
Thanks @marcocitus… there was no intention of `pg_shard` honoring `DISTRIBUTE BY APPEND` at the moment anyways, as it should only understand hash partitioning. I'll work to ensure it understands there are tables that are distributed but which it should not touch.

As far as caching goes… bleh. I forgot we had added caching. `pg_shard` doesn't expect things to change underneath it. It doesn't cache empty results, so if _no_ shards exist you shouldn't see caching. But once shards exist we currently expect them to be long-lived (placements, on the other hand, can change state, etc.)

Do we anticipate having `\STAGE` work with `pg_shard` anytime soon? `pg_shard` assumes partitions are that: partitions.

ozgune:
I had a quick clarification question.

With these changes, which migration scenario do we intend to handle? (a) pg_shard -> CitusDB, (b) CitusDB -> pg_shard (shard rebalancer?), or (c) CitusDB <-> pg_shard

CitusDB currently observes state changes when we \stage to a new shard (append partitioned), append to an existing shard, or rebalance a table. pg_shard observes state changes when we create shards for a new table, or fail to write to a shard placement (hash partitioned).

Two other questions -- don't know, just asking. Who has the authoritative metadata, CitusDB or pg_shard with these changes? Under what conditions does the authority change?

jasonmp85:
Because this PR actually causes `pg_shard` to write through to CitusDB when CitusDB is present, there is only ever one actual metadata store. When using CitusDB, CitusDB's `pg_dist*` tables store metadata. When using `pg_shard` with PostgreSQL, it uses its own tables.

That means the only possibility is _stale_ metadata (i.e. from caching). So we'll want to address that.

jasonmp85:
@ozgune what do you mean "migration scenario"? I can see:

  * Using `pg_shard` with PostgreSQL and upgrading your underlying DB to CitusDB
  * Using CitusDB without `pg_shard` and wanting to add it for `INSERT`s
  * Using `pg_shard` to create new distributed tables within CitusDB and have it know how to query them

Am I missing any other possibilities?

ozgune:
@jasonmp85 Yup, I was curious more about the use-cases, where we needed to sync metadata between pg_shard and CitusDB.

1. Using pg_shard with PostgreSQL and upgrading your underlying DB to CitusDB (pg_shard metadata tables -> CitusDB metadata tables)
2. Using pg_shard to create new distributed tables within CitusDB and have it know how to query them (pg_shard -> CitusDB)
3. Using CitusDB without pg_shard and wanting to add it for INSERTs: In this case, is the CitusDB table partitioned by append? Could we overlook this scenario by guiding CitusDB users to start with pg_shard?
4. Using pg_shard for Inserts, CitusDB for Selects, and shard rebalancer to rebalance the shards (pg_shard -> CitusDB, then CitusDB -> pg_shard): Is adding code to the shard rebalancer an option to sync the CitusDB metadata back to pg_shard?

I'm wonder if we can simplify the problem down to (pg_shard -> CitusDB). **If** we could and wanted to, could we then put a trigger on the pg_shard metadata to propagate the update to CitusDB system catalogs?

jasonmp85:
Let's stop using the word sync. There is no syncing here.

jasonmp85:
To clarify, everything is always in sync… there is no sync step. `pg_shard` will directly write to CitusDB metadata tables when installed within CitusDB, and the right things are in place to permit a migration from PostgreSQL + `pg_shard` to CitusDB using `pg_dump` to write out a `pg_shard` master and turn it into a `CitusDB` master.

jasonmp85:
By the way, the migration step is to write out all normal tables _except_ the `pg_shard` metadata using `pg_dump -N pgs_distribution_metadata`. You apply that to a fresh CitusDB install that will be your master .Then you dump the `pg_shard` metadata with `pg_dump --inserts -t 'pgs_distribution_metadata.*'` and apply _that_ to the new CitusDB master. From that point onwards all previous `pg_shard` tables will be seen by CitusDB, and all `pg_shard` and CitusDB metadata will be identical.

jasonmp85:
If you're using `pg_shard` within CitusDB from square one, it _always_ remains in sync: it sees all CitusDB tables, and CitusDB sees all `pg_shard` metadata.

jasonmp85:
I'm still figuring out how I want to test the CitusDB-specific parts of this, but the main code is complete.

distribution_metadata.c
```diff
@@ -584,6 +780,27 @@ LoadShardIntervalRow(int64 shardId, Oid *relationId, char **minValue,
 
 
 /*
+ * NextSequenceId allocates and returns a new unique id generated from the given
+ * sequence name.
+ */
+static uint64
+NextSequenceId(char *sequenceName)
```
sumedhpathak: Should we pass the schema name as an argument to this function?

distribution_metadata.c
```diff
@@ -594,19 +811,32 @@ TupleToShardPlacement(HeapTuple heapTuple, TupleDesc tupleDescriptor)
...
+	if (UseCitusMetadata)
+	{
+		Datum shardLengthDatum = heap_getattr(heapTuple, AttrNumShardPlacementShardLength,
+											  tupleDescriptor, &isNull);
+		shardPlacement->id = (int64) DatumGetObjectId(idDatum);
+		shardPlacement->shardLength = DatumGetInt64(shardLengthDatum);
+	}
+	else
+	{
+		shardPlacement->id = DatumGetInt64(idDatum);
```
sumedhpathak: Is there a reason we don't cast to int64 here (but do so above)?

pg_shard--1.2.sql
```diff
@@ -1,53 +1,140 @@
...
+AS 'MODULE_PATHNAME'
+LANGUAGE C STABLE STRICT;
+
+DO $$
+DECLARE
+    replication_factor integer;
+    use_citus_metadata boolean := false;
+BEGIN
+    BEGIN
+        replication_factor = current_setting('shard_replication_factor')::integer;
```
sumedhpathak: Can we detect based on existence of pg_dist_partition itself? That seems like a more logical way to detect presence of existing metadata tables, and thus creating views to point to it.

distribution_metadata.c
jasonmp85: We don't pass it into any other functions, and in the previous implementation of this function (essentially unchanged here) it was just a `#define` we passed into `makeRangeVar` or whatnot, so I don't see why it needs to be exposed: this function is for getting sequence values related to metadata (see the previous implementation, which was locked to the metadata schema), so it should encapsulate the knowledge of where they live.

distribution_metadata.c
jasonmp85: Yes. Citus does not _have_ placement IDs: it just uses an `Oid`. `Oid` is essentially `int32`, so I have an explicit cast to show that we're using the wider `int64` in the in-memory `struct` representation.

pg_shard--1.2.sql
jasonmp85: In this particular place: yes, we could. The issue is in `_PG_init`, which runs on the postmaster and not in any particular backend. Because of that, it doesn't have access to the caches typically used to check for relations, catalogs, etc.

I kept the check in the SQL identical for consistency purposes, but if you think it's clearer to change back I can do that.

pg_shard--1.2.sql
```diff
@@ -1,53 +1,140 @@
...
+
+            CREATE VIEW shard_placement AS
+                SELECT oid::bigint AS id,
+                       shardid     AS shard_id,
+                       shardstate  AS shard_state,
+                       nodename    AS node_name,
+                       nodeport    AS node_port
+                FROM   pg_dist_shard_placement
+
+            CREATE TRIGGER shard_placement_insert INSTEAD OF INSERT ON shard_placement
```
sumedhpathak: Will deletes/updates error out? Or should we have triggers similar to the insert.

pg_shard--1.2.sql
```diff
@@ -1,53 +1,140 @@
...
+BEGIN
+    BEGIN
+        replication_factor = current_setting('shard_replication_factor')::integer;
+        use_citus_metadata = (replication_factor IS NOT NULL);
+    EXCEPTION
+        WHEN undefined_object THEN
+            use_citus_metadata = false;
+    END;
+
+    IF use_citus_metadata THEN
```
sumedhpathak: These functions mirror what we're doing in code as well. It seems wasteful to have similar logic replicated across both. If the SPI machinery would allow us to keep this logic only here (and keep the C metadata functions simpler), then I'd prefer that. Of course, that is predicated upon whether using SPI causes other locking issues (or is prohibitive due to other reasons).
It will not match Citus metadata functions, but I am OK with that. @ozgune Any concerns there?

pg_shard--1.2.sql
jasonmp85: They "Just Work™".

pg_shard--1.2.sql
jasonmp85: Ack. Switched to SPI.

pg_shard--1.2.sql
jasonmp85: Switching to what you said.

distribution_metadata.c
jasonmp85: Function no longer exists; no problem.

distribution_metadata.c
jasonmp85: Entirely encapsulated within the new Create/Update/Delete functions, now.

pg_shard--1.1--1.2.sql
```diff
@@ -1,13 +1,242 @@
...
+
+		CREATE FUNCTION adapt_and_insert_partition() RETURNS trigger AS $aaip$
+		BEGIN
+			INSERT INTO pg_dist_partition
+						(logicalrelid,
+						 partmethod,
+						 partkey)
+			VALUES      (NEW.relation_id,
+						 NEW.partition_method,
+						 column_name_to_column(NEW.relation_id, NEW.key));
```
onderkalaci: @jasonmp85 We need to add explicit cast at this line. My tests with pg_dump showed me that we need to do following:
```SQL
column_name_to_column(NEW.relation_id::Oid, NEW.key));
```

The reason is that pg_dump dumps it as int (at least with no special configuration), so we see an error here.

jasonmp85: Actually, I think the column type of the view and table should just be `regclass`. It will solve that.

distribution_metadata.c
```diff
@@ -389,7 +391,7 @@ IsDistributedTable(Oid tableId)
 	 * The query below hits the partition metadata table, so if we don't detect
 	 * that and short-circuit, we'll get infinite recursion in the planner.
 	 */
-	if (tableId == partitionMetadataTableOid)
+	if (IsSystemNamespace(tableNamespaceOid) || tableId == partitionMetadataTableOid)
```
jasonmp85: Note to @sumedhpathak: the view-based `SELECT` gets rewritten with a subselect that hits the CitusDB metadata directly. So the recursion issue re-emerges. I know in #110 you liked me testing against the table `Oid` rather than the namespace of the target table, but I think it's clear that no catalog table ("system namespace") is ever distributed.

So adding this small check here fixes the CitusDB case.

pg_shard.c
```diff
@@ -579,7 +579,7 @@ ExtractFirstDistributedTableId(Query *query)
 	{
 		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) lfirst(rangeTableCell);
 
-		if (IsDistributedTable(rangeTableEntry->relid))
+		if (rangeTableEntry->rtekind == RTE_RELATION && IsDistributedTable(rangeTableEntry->relid))
```
jasonmp85: Note to @sumedhpathak: According to comments for `RangeTblEntry`, `relid` is only set for range tables whose kind is `RTE_RELATION`. The views rewrite to subqueries, which have a `relid` of zero, which we happily pass to `IsDistributedTable`. By adding this check we avoid that.

If the subqueries themselves contain distributed relations, that's OK, because the walker will have grabbed their entries and we'll come across them in this loop anyhow.

distribution_metadata.c
sumedhpathak: I see, so there are two use-cases which hits the recursion use-case: queries against the CitusDB partition table as well as queries against the pg_shard partition table? The clause captures the CitusDB case, and the second captures the pg_shard case?

If so, could we modify the comment to make that clear?

pg_shard.c
sumedhpathak: What is the impact of passing relid to 0 to IsDistributedTable? Does it crash? I'd think it would be OK to pass 0 as that should just return false?

- Sumedh

pg_shard--1.1--1.2.sql
```diff
@@ -1,13 +1,242 @@
...
+			PERFORM sync_table_metadata_to_citus(relation_id::regclass);
+		END LOOP;
+
+		DROP TABLE pgs_distribution_metadata.partition,
+				   pgs_distribution_metadata.shard,
+				   pgs_distribution_metadata.shard_placement;
+
+		CREATE FUNCTION adapt_and_insert_shard() RETURNS TRIGGER AS $aais$
+		BEGIN
+			IF NEW.id IS NULL THEN
```
sumedhpathak: Should we throw an exception if new.id isn't null? Is there a use-case we expect it wouldnt be NULL?

pg_shard--1.1--1.2.sql
```diff
@@ -1,13 +1,242 @@
...
+	BEGIN
+		RAISE WARNING '%', warning_msg;
+
+		-- grab lock to ensure single writer for upsert
+		LOCK TABLE pg_dist_shard_placement IN EXCLUSIVE MODE;
+
+		-- First, update the health of shard placement rows already copied
+		-- from pg_shard to CitusDB. Health is the only mutable attribute,
+		-- so it is presently the only one needing the UPDATE treatment.
+		UPDATE pg_dist_shard_placement
```
sumedhpathak: What about nodename/nodeport? If someone ran the shard-repair functionality, those values could change too?

pg_shard--1.1--1.2.sql
```diff
@@ -1,13 +1,242 @@
...
+			   shard_state,
+			   dummy_shard_length,
+			   node_name,
+			   node_port
+		FROM   pgs_distribution_metadata.shard_placement
+			   LEFT OUTER JOIN pg_dist_shard_placement
+							ON ( shardid = shard_placement.shard_id AND
+								 nodename = shard_placement.node_name AND
+								 nodeport = shard_placement.node_port )
+		WHERE  shardid IS NULL AND
```
sumedhpathak: Could we add aliases (like pg_shard_table.shardid / citusdb_table.shard_id)? Makes it a bit easier to grok.

pg_shard--1.1--1.2.sql
sumedhpathak: Don't I really want to delete everything in CitusDB's tables for that relationId, and then re-sync?

pg_shard--1.1--1.2.sql
```diff
@@ -1,13 +1,242 @@
...
+			   storage,
+			   min_value,
+			   max_value
+		FROM   pgs_distribution_metadata.shard
+			   LEFT OUTER JOIN pg_dist_shard
+							ON ( shardid = shard.id )
+		WHERE  shardid IS NULL AND
+			   relation_id = table_relation_id;
+
+		-- Finally, copy pg_shard partition rows not yet in CitusDB's metadata
```
sumedhpathak: For the given relationId, there should be just one row right? The comment and the query are a bit confusing. Can we modify the query to denote that we expect only one row in this table for the given relationId?

pg_shard--1.2.sql
```diff
@@ -3,57 +3,172 @@
...
+				FOR EACH ROW
+				EXECUTE PROCEDURE adapt_and_insert_shard_placement()
+
+			CREATE VIEW partition AS
+				SELECT logicalrelid AS relation_id,
+					   partmethod   AS partition_method,
+					   column_to_column_name(logicalrelid, partkey) AS key
+				FROM   pg_dist_partition
+
+			CREATE TRIGGER partition_insert INSTEAD OF INSERT ON partition
```
sumedhpathak: We trigger on INSERT's. Will updates/deletes error out? 

If so, do we need some check in the shard-repair functionality that errors if its on CitusDB (with message to use the shard rebalancer maybe)?

pg_shard--1.1--1.2.sql
jasonmp85: But I'll add the `Oid` cast for this code review.

distribution_metadata.c
jasonmp85: Well, they're both queries against the relation named `pgs_distribution_metadata.partition`. That relation is either a view into the CitusDB tables or a normal table. In the view case, the query gets rewritten into a subquery like `SELECT NULL FROM (SELECT logicalrelid AS relation_id ... FROM pg_dist_partition) WHERE relation_id = $1`. Our planner sees the reference to `pg_dist_partition` in the subquery and we hit recursion there without this check.

IMO it's still probably more consistent to just check whether the table is in `pgs_distribution_metadata` or in a system catalog, but in an earlier comment you asked me to check for the partition table `oid` directly. But since we now also have to handle a CitusDB catalog, and since I don't want to have the C code know about the `oid`s of CitusDB tables, I check for whether the target table is a catalog table (which works in CitusDB).

As for the comment; clearly it needs some clarification :grin:. Updating. (done)

pg_shard.c
jasonmp85: `0` is the same as `InvalidOid`. Right now I think the code "just works" and vacuously returns `false`, but we pass the `oid` argument to `get_rel_namespace`, so we're now making an assumption that it will not barf on `InvalidOid` input.

A bigger issue is: can we trust that it's actually zero for non-relations? What if somewhere in the PostgreSQL code `palloc` is used instead of `palloc0`? Then we've got a bogus relation identifier being passed to this function, which could have undefined behavior.

After reasoning this out here, I'm thinking either `IsDistributedTable` should just be more explicit in its handling of `InvalidOid`: it should error out if we aren't going to handle it, or return early if we will.

Since I do like the idea of allowing it to simply handle `InvalidOid`, I'll just have it short-circuit if it sees that argument. Sound OK? (done)

pg_shard--1.1--1.2.sql
jasonmp85: The `INSERT` statements produced by `pg_dump` might have a specified identifier, or the user might have their own UDFs that expect to be able to insert rows into the metadata tables.

It's more code to error out on non-`NULL`, and it's more surprising to the end user. Might as well just do what's expected and allow general `INSERT` statements.

pg_shard--1.1--1.2.sql
jasonmp85: This function already exists in `pg_shard` and is not changing in this release. The only change here is to add the `RAISE WARNING` as users do not need to call this method any longer now that metadata just stays in sync.

In v1.3 we'll remove this function entirely, but we need it in v1.2 to aid in migration from 1.1 to 1.2, since CitusDB + `pg_shard` users need this workflow to upgrade:

  * Perform one last sync from `pg_shard` to CitusDB
  * Drop `pg_shard` metadata tables
  * Recreate `pg_shard` metadata relations as views into CitusDB
  * Add warning to sync method to let users know they no longer need it

_Technically_ we could remove this function altogether right now as the last step, but if we're being strict about version numbers that's a breaking change and would necessitate a bump to v2.0. So we'll have a deprecation warning until then.

jasonmp85: This function only syncs from `pg_shard` to CitusDB using an upsert. Using the rebalancer or otherwise modifying the CitusDB column values is not a supported use case. And none of `pg_shard`'s code changes the `nodename` or `nodeport` of an existing placement (they only change state).

So the only thing this method handled was changing placement state or adding new tables, shards, and placements. It lets CitusDB know about `pg_shard`, not anything else.

pg_shard--1.1--1.2.sql
jasonmp85: Again, since this method does not change (apart from the deprecation warning) in this version, I'd prefer to leave it alone. In fact, there is no reason to change it at all anymore, since it's no longer needed.

pg_shard--1.1--1.2.sql
jasonmp85: Gah. Seems like you spent more time code reviewing the metadata sync function we already shipped in v1.1. I should have been more clear about why I redefined it in this upgrade file. It's only to add a deprecation warning. I do not intend to change it at all until it is dropped during v2.0.

pg_shard--1.2.sql
jasonmp85: I believe `UPDATE` and `DELETE` will "just work" using the PostgreSQL updatable views logic where possible, and where not will issue an appropriate error. I'll check it out and get back to you about it.

jasonmp85:
Code rebased on latest develop, tests fixed, and changes from feedback made (denoted by _(done)_ in one of my replies).

@sumedhpathak, can you see what else needs attention on this PR?

PR #104, opened 2015-04-27
Bugfix deparse constraint#94

### LINKED ISSUE #94: Table constraints deparsed incorrectly
While reviewing error message style in `pg_shard`, I noticed that an error message in [the DDL extension code][1] referred to a set of constraints as a _column constraint_. "That's odd", I thought… "I thought those were table constraints?"

The unit test for this section of code extends the following DDL command:

```sql
CREATE TABLE employees ( 
    first_name text NOT NULL, 
    last_name  text NOT NULL, 
    id         bigint NOT NULL, 
    salary     numeric DEFAULT 0.00, 
    start_date timestamp without time zone, 
    resume     text, 
    CONSTRAINT sal_check CHECK (salary >= 0.00) 
);
```

There are several column constraints here (specifically, `NOT NULL` ones), but the final `CONSTRAINT` is a table constraint because it is not syntactically attached to any column. But here's what the extended DDL looks like:

```sql
CREATE TABLE employees_12345 ( 
    first_name  text NOT NULL, 
    last_name   text NOT NULL, 
    id          bigint NOT NULL, 
    salary      numeric DEFAULT 0.00, 
    start_date  timestamp without time zone, 
    resume text CONSTRAINT sal_check CHECK (salary >= 0.00) 
);
```

The constraint is generated _right next to the final column_, when in fact there should be a comma to separate it. The unit test itself is wrong (it should have a comma).

[1]: https://github.com/citusdata/pg_shard/blob/v1.1.0/extend_ddl_commands.c?ts=4#L628-L662

onderkalaci:
Fixes #94

### DIFF
--- a/extend_ddl_commands.c
+++ b/extend_ddl_commands.c
@@ -641,7 +641,7 @@ DeparseCreateStmt(CreateStmt *createStmt, Oid masterRelationId)
 							errdetail("Only the CHECK form is supported.")));
 		}
 
-		if (firstConstraintPrinted)
+		if (firstAttributePrinted || firstConstraintPrinted)
 		{
 			appendStringInfoString(deparsedCreate, ", ");
 		}
--- a/extend_ddl_commands.c
+++ b/extend_ddl_commands.c
@@ -684,7 +684,7 @@ DeparseCreateStmt(CreateStmt *createStmt, Oid masterRelationId)
 
 /*
  * TableElementsOfType walks over the table element list and returns only those
- * nodes who's type matches the given node type.
+ * nodes whose type matches the given node type.
  */
 static List *
 TableElementsOfType(List *tableElementList, NodeTag nodeType)
--- a/expected/extend_ddl_commands.out
+++ b/expected/extend_ddl_commands.out
@@ -28,9 +28,9 @@ SELECT extend_ddl_command('employees', 12345, 'CREATE TABLE employees (first_nam
 						  'NULL, salary numeric DEFAULT 0.00, start_date timestamp ' ||
 						  'without time zone, resume text, CONSTRAINT sal_check ' ||
 						  'CHECK (salary >= 0.00))');
-                                                                                                         extend_ddl_command                                                                                                         
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


onderkalaci:
@jasonmp85 After our chat, I understand the bug.
Yes, as you suggest, we already find all check constraints(column and table) in the table and convert them to table constraints. The following quote from the PostgreSQL documentation also reinforces our approach:

```We say that the first two constraints are column constraints, whereas the third one is a table constraint because it is written separately from any one column definition. Column constraints can also be written as table constraints, while the reverse is not necessarily possible, since a column constraint is supposed to refer to only the column it is attached to. (PostgreSQL doesn't enforce that rule, but you should follow it if you want your table definitions to work with other database systems.) ```

This commit fixes [DeparseCreateStmt](https://github.com/citusdata/pg_shard/blob/develop/extend_ddl_commands.c#L644) function, in which we fail to add a comma for the first table constraint that is deparsed. Since we forget the first comma after column definitions, the first table constraint was considered as column constraint.

Basically, I tried to use the same approach in [pg_shard_get_tableschemadef_string](https://github.com/citusdata/pg_shard/blob/develop/generate_ddl_commands.c#L301), which works as expected.

File: extend_ddl_commands.c [hunk 1]
```diff
@@ -641,7 +641,7 @@ DeparseCreateStmt(CreateStmt *createStmt, Oid masterRelationId)
 							errdetail("Only the CHECK form is supported.")));
 		}
 
-		if (firstConstraintPrinted)
+		if (firstAttributePrinted || firstConstraintPrinted)
```
jasonmp85: At first I thought maybe this was unnecessary, and perhaps the code could be simplified by always printing a comma before each constraint. After all, shouldn't we have printed a column by the time we get here?

Then I remembered I frequently declare zero-column tables in testing, and wondered whether I could have a table with zero columns but still some constraints:

```sql
CREATE TABLE empty_with_constraint (CHECK (1 > 2));
-- Table "public.empty_with_constraint"
-- ┌────────┬──────┬───────────┐
-- │ Column │ Type │ Modifiers │
-- ├────────┼──────┼───────────┤
-- └────────┴──────┴───────────┘
-- Check constraints:
--     "empty_with_constraint_check" CHECK (1 > 2)
```

This is of course insane, but it shows that the check you've written is necessary.

jasonmp85:
Looks good. Merging this!

:shipit:

PR #105, opened 2015-04-27
Check column name length in ColumnNameToColumn

jasonmp85:
Discovered that system cache lookups have Asserts that fail if any name over `NAMEDATALEN` is provided. This change gives a nicer behavior than just crashing the whole backend.

I also discovered that `ResolveRelationId` silently truncates its input, which I initially thought I'd fix. It appears _many_ functions that use the [`name` data type](http://www.postgresql.org/docs/9.4/static/datatype-character.html#DATATYPE-CHARACTER-SPECIAL-TABLE) also silently truncate their input, so I left that as-is. At any rate, it accepts a qualified name (delimited by periods), so detecting whether any given part of that is too long is more trouble than it's worth, especially when other pieces of PostgreSQL proper truncate regardless.

### DIFF
--- a/distribution_metadata.c
+++ b/distribution_metadata.c
@@ -436,12 +436,14 @@ DistributedTablesExist(void)
 /*
  * ColumnNameToColumn accepts a relation identifier and column name and returns
  * a Var that represents that column in that relation. This function throws an
- * error if the column doesn't exist or is a system column.
+ * error if the column doesn't exist, is a system column, or if the provided
+ * name is longer than NAMEDATALEN bytes.
  */
 Var *
 ColumnNameToColumn(Oid relationId, char *columnName)
 {
 	Var *partitionColumn = NULL;
+	AttrNumber columnId = InvalidAttrNumber;
 	Oid columnTypeOid = InvalidOid;
 	int32 columnTypeMod = -1;
 	Oid columnCollationOid = InvalidOid;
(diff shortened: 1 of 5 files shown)


distribution_metadata.c
```diff
@@ -450,7 +452,15 @@ ColumnNameToColumn(Oid relationId, char *columnName)
 	const Index tableId = 1;
 	const Index columnLevelsUp = 0;
 
-	AttrNumber columnId = get_attnum(relationId, columnName);
+	/* check length to prevent failed Assert for long names in get_attnum */
+	if (strnlen(columnName, NAMEDATALEN) == NAMEDATALEN)
+	{
+		ereport(ERROR, (errcode(ERRCODE_NAME_TOO_LONG), errmsg("columnName too long"),
```
onderkalaci: What do you think changing error message to "column name is too long"

distribution_metadata.c
```diff
@@ -450,7 +452,15 @@ ColumnNameToColumn(Oid relationId, char *columnName)
...
+						errdetail("columnName must be less than %d characters.",
```
onderkalaci: Similarly, "column name must be less ..."

onderkalaci:
Hey @jasonmp85,

I looked and tested the change.  I have a tiny comment on the error message. If that makes sense, we can change it. Or, we can just leave it as it is, and ship it.

distribution_metadata.c
jasonmp85: This made me double-check other error messages for variable names. I found a few places, all of them [in this region](https://github.com/citusdata/pg_shard/blob/d1a335dae59413b2db47896d69b4dab7ac1b1224/create_shards.c#L188-L222).

In that case, the messages are actually referring to user-specified function arguments, so I think that's an argument to actually change them to match the underscore-separated variable names used in the SQL. So I'll do that.

For here, though… this is an internal C method and the error is about calling with bad C parameters. So it's more of an internal error. That would be a case to make this an `Assert`, but this could be triggered by someone simply updating values in the partition table using SQL.

That might be a case to redefine that column as the `name` type rather than `text`, but in lieu of that I'll check other PostgreSQL code to see whether any internal functions complain using their own parameter names even though the end-user might not even know what those variables are talking about.

Otherwise, I might do away with the _column_ word altogether and see what kind of errors I can dig up that reference names being too long.

jasonmp85: OK, agreed and changing.

distribution_metadata.c
jasonmp85: Done.

Issue #108, opened 2015-05-05
Support Function Calls for Sharded Modification.


gmcquillan:
I'm experimenting with using HyperLogLog (HLL) data types in some columns. One problem with these is that they take up quite a lot more space than a BIGINT. pg_shard potentially allays a lot of those issues. The extension that provides the HLL datatype is [this one](https://github.com/aggregateknowledge/postgresql-hll) by aggregateknowledge. These two extensions seem complimentary for warehousing purposes.

Surprisingly, these data types work with sharded tables for most types of reads, but not for writes (see below). When I attempt an update like so:

```sql
update test_hll_shard set users = users||hll_hash_text('foobar') where date = date('2015-01-08');
```
I get: 
```
ERROR:  cannot plan sharded modification containing values which are not constants or constant expressions
```

 I can sort of work around this by setting the literal bytes in this field. Which works fine -- but adding HLL values requires a read and a write. Since pg_shard (understandably) doesn't allow for more than a single statement transaction, this leaves my use-case vulnerable to race conditions in multi-writer environments.

Since this function is available on the workers, and is deterministic based on the value of the existing row and the new HLL value to be added, there shouldn't be any issue with dispatching this expression through to the workers. 

Is there a *hard limitation* preventing pg_shard from dispatching modifications for non-constant expressions?

gmcquillan:
Similarly, as mentioned in the pg_shard-users list, subqueries might be another work around to the function issue, but they are also not supported yet.

jasonmp85:
> Since this function is available on the workers, and is deterministic based on the value of the existing row and the new HLL value to be added, there shouldn't be any issue with dispatching this expression through to the workers.

Are you sure this issue is due to the `hll_hash_text` function and not the `date` one? The HLL function is marked as `IMMUTABLE`, which means it should be folded into a constant before being pushed to the workers…

This might be an argument that the error message should be clarified to say _which_ expression is non-constant.

jasonmp85:
> Is there a hard limitation preventing pg_shard from dispatching modifications for non-constant expressions?

At the moment, yes. `pg_shard` performs `UPDATE`s by grabbing a lock for a shard and updating all replicas of that shard. This means there is no first-class replication happening but rather a replica-by-replica execution of the query at hand. If that query had e.g. a `now()` call, the output might differ among replicas, resulting in data divergence.

So unless we can fold an expression into a constant expression, modifications are unsafe if replication is in play.

gmcquillan:
[...]

PR #109, opened 2015-05-07
Conditionally support functions

gmcquillan:
Addresses issue #108. 

Mostly submitting to show that my idle curiosity is satisfied. Naturally, allowing function calls to go through to the shard workers includes a lot of unknowns, particularly around which extensions are loaded, which stored procedures have been saved, probably other things that I don't know about off the top of my head. I'm not sure what the best way to handle that would be, but maybe allowing a whitelist of function names that the maintainer of a cluster knows is on each of the postgres workers would be a good start? For now, this patch satisfies my use-case.

- Tested on Postgres 9.4 on Ubuntu 15.04, and Ubuntu 14.04.
- Supplying an environment variable during compile time ``SUPPORT_SHARD_EXPRS`` set to 'yes' will result in the preprocessor to remove the checks for expressions being passed to shard workers from the query tree.

Example Table Definition:

```sql
template1=# \d test
     Table "public.test"
 Column |  Type   | Modifiers 
--------+---------+-----------
 id     | integer | 
 date   | date    | 
```

Example before patch:

```sql
template1=# insert into test values(111223232, now()::date);
ERROR:  cannot plan sharded modification containing values which are not constants or constant expressions
```
Example after patch:

```sql
template1=# insert into test values(111223233, now()::date);
INSERT 0 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ OBJS = connection.o create_shards.o citus_metadata_sync.o distribution_metadata.
 	   extend_ddl_commands.o generate_ddl_commands.o pg_shard.o prune_shard_list.o \
 	   repair_shards.o ruleutils.o
 
-PG_CPPFLAGS = -std=c99 -Wall -Wextra -I$(libpq_srcdir)
+PG_CPPFLAGS = -std=c99 -Wall -Wextra -I$(libpq_srcdir) -DEXP_SUPPORT_EXPRS
 
 # pg_shard and CitusDB have several functions that share the same name. When we
 # link pg_shard against CitusDB on Linux, the loader resolves to the CitusDB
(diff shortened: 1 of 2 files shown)


jasonmp85:
Unfortunately, those checks are crucial to ensure that each replica has exactly the same data set. We are exploring options here (more first-class replication using something like BDR, or folding some functions on the master, like `now`), but since we're doing modifications to replicas one-at-a-time, it is unsafe to evaluate anything except `IMMUTABLE` functions during modifications.

gmcquillan:
Makes sense. It was mostly just curiosity that I sent this patch. I'll go ahead and close to keep the clutter out of the PR queue.

PR #110, opened 2015-05-08
Switch to SPI for metadata functions

jasonmp85:
In #103, we discovered that using low-level operations is kind of unwieldy when the code needs to switch between different backing stores. By using SPI, we can overlay a `VIEW` on the CitusDB metadata catalogs and have everything "just work".

This change prepares for that by switching to SPI for our metadata functions. The functions will now respect any `VIEW` or `TRIGGER` we need to adapt to CitusDB's underlying representation.


create_shards.c
```diff
@@ -256,6 +253,15 @@ master_create_worker_shards(PG_FUNCTION_ARGS)
 			shardMaxHashToken = INT_MAX;
 		}
 
+		/* insert the shard metadata row along with its min/max values */
+		minHashTokenText = IntegerToText(shardMinHashToken);
+		maxHashTokenText = IntegerToText(shardMaxHashToken);
+		shardId = CreateShardRow(distributedTableId, shardStorageType, minHashTokenText,
```
sumedhpathak: We seem to have switched to associating the id column with its sequence, with an auto-increment. Any reason why? 
This currently switches the order of creation of shard and shard placement. We initially created all the placements first, and then created the shard.

sumedhpathak: We also error out if we were unable to create placements. Thus creating the shard row after that was considered safe. Will the shard row we've created get rolled back if we error out on placement creation? (See line 296 for the error)

create_shards.c
```diff
@@ -236,14 +236,11 @@ master_create_worker_shards(PG_FUNCTION_ARGS)
 
 	for (shardIndex = 0; shardIndex < shardCount; shardIndex++)
 	{
-		uint64 shardId = NextSequenceId(SHARD_ID_SEQUENCE_NAME);
+		int64 shardId = -1;
```
sumedhpathak: I initially thought we had shardId's defined as unsigned everywhere, but that doesn't seem to be the case. Do you know why NextSequenceId was returning uint64's, but shardId's were int64's everywhere? Do we have other places with this inconsistency?

distribution_metadata.c
```diff
@@ -126,48 +115,36 @@ List *
...
+	spiStatus = SPI_execute_with_args("SELECT s.id, s.relation_id, s.min_value, "
+									  "s.max_value, p.partition_method, p.key "
+									  "FROM pgs_distribution_metadata.shard AS s "
+									  "JOIN pgs_distribution_metadata.partition AS p "
+									  "ON s.relation_id = p.relation_id "
+									  "WHERE s.relation_id = $1", argCount, argTypes,
+									  argValues, NULL, false, 0);
+	Assert(spiStatus == SPI_OK_SELECT);
+
+	oldContext = MemoryContextSwitchTo(upperContext);
```
sumedhpathak: We set upperContext = CurrentMemoryContext, so what we switching from and to? Aren't we already in the current memory context?

distribution_metadata.c
```diff
@@ -524,62 +467,68 @@ ColumnNameToColumn(Oid relationId, char *columnName)
...
-	indexRelation = relation_openrv(indexRangeVar, AccessShareLock);
+	ShardInterval *shardInterval = NULL;
+	bool isNull = false;
+	Oid intervalTypeId = InvalidOid;
+	int32 intervalTypeMod = -1;
+	Oid inputFunctionId = InvalidOid;
+	Oid typeIoParam = InvalidOid;
 
-	ScanKeyInit(&scanKey[0], 1, BTEqualStrategyNumber, F_INT8EQ, Int64GetDatum(shardId));
+	Datum idDatum = SPI_getbinval(heapTuple, tupleDescriptor, 1, &isNull);
```
sumedhpathak: The number is presumably the field location? This makes me a bit uncomfortable, since we are coupled to the way the query is defined in a different function. The additional complication is the query also joins with the partition table to get the partition column. How can we make the two consistent?

For one, I'd rather the query just returns the fields of the table. Maybe we could then use the relation tupleDescriptor (instead of query), so we know what this function expects?  

Is there a downside to sticking with the PartitionColumn function to load the column and its type like we did before? (Instead of joining with partition).

distribution_metadata.c
```diff
@@ -182,47 +159,37 @@ ShardInterval *
...
-
-	/* first read the related row from the shard table */
-	LoadShardIntervalRow(shardId, &relationId, &minValueString, &maxValueString);
-
-	/* then find min/max values' actual types */
-	partitionType = PartitionType(relationId);
-	if (partitionType == HASH_PARTITION_TYPE)
+	Oid argTypes[] = { INT8OID };
+	Datum argValues[] = { Int64GetDatum(shardId) };
+	const int argCount = sizeof(argValues) / sizeof(argValues[0]);
```
sumedhpathak: Do we need to do this? It feels like we could just mark it as 1 no?

distribution_metadata.c
```diff
@@ -419,28 +370,30 @@ PartitionType(Oid distributedTableId)
...
-	heapRangeVar = makeRangeVar(METADATA_SCHEMA_NAME, PARTITION_TABLE_NAME, -1);
-	heapRelation = relation_openrv(heapRangeVar, AccessShareLock);
+	Oid argTypes[] = { OIDOID };
+	Datum argValues[] = { ObjectIdGetDatum(tableId) };
+	const int argCount = sizeof(argValues) / sizeof(argValues[0]);
+	int spiStatus = 0;
 
-	ScanKeyInit(&scanKey[0], ATTR_NUM_PARTITION_RELATION_ID, InvalidStrategy,
-				F_OIDEQ, ObjectIdGetDatum(tableId));
+	if (tableNamespaceOid == metadataNamespaceOid)
```
sumedhpathak: What is this check for? Does this mean if the table is *in* in the ``pgs_distribution_metadata`` namespace, it isn't considered distributed?

distribution_metadata.c
```diff
@@ -829,22 +744,15 @@ LockShard(int64 shardId, LOCKMODE lockMode)
...
 	uint32 keyUpperHalf = (uint32) (shardId >> 32);
 	uint32 keyLowerHalf = (uint32) shardId;
+	bool sessionLock = false;   /* we want a transaction lock */
+	bool dontWait = false;      /* block indefinitely until acquired */
 
 	LOCKTAG lockTag;
 	memset(&lockTag, 0, sizeof(LOCKTAG));
 
-	SET_LOCKTAG_ADVISORY(lockTag, MyDatabaseId, keyUpperHalf, keyLowerHalf, 0);
+	Assert(lockMode == ExclusiveLock || lockMode == ShareLock);
```
sumedhpathak: Is this change related to the SPI one? Any reason to convert it from an error to Assert?

repair_shards.c
```diff
@@ -137,10 +137,7 @@ master_copy_shard_placement(PG_FUNCTION_ARGS)
 	}
 
 	/* the placement is repaired, so return to finalized state */
-	DeleteShardPlacementRow(targetPlacement->id);
```
sumedhpathak: Is there a reason to diverge from this pattern? I realize we now need to do so because of the auto-incrementing sequence. Are there any advantages/disadvantages vs the delete/insert vs update pattern? Maybe we loop in @ozgune on this?

jasonmp85: Yes. CitusDB has fields `pg_shard` doesn't know about (shard length or whatnot). If we do `DELETE`/`INSERT`, `pg_shard` must know about them. If we just do an `UPDATE`, we can update the fields we care about without having to keep track of and re-insert the ones we don't.

distribution_metadata.c
jasonmp85: Unrelated. Was just cleaning up. We use `ereport` for stuff the user might conceivably trigger. If we call this method with an incorrect lock mode, that's _our_ problem, so an `Assert` makes much more sense here.

create_shards.c
jasonmp85: > We seem to have switched to associating the id column with its sequence, with an auto-increment. Any reason why?

By using a default value, the C code can remain ignorant of the names of sequences used by CitusDB or pg_shard. This means the only place that knowledge lives will be in the SQL install script: the C code will not have any CitusDB/pg_shard detection code for metadata functionality.

> We also error out if we were unable to create placements. Thus creating the shard row after that was considered safe. Will the shard row we've created get rolled back if we error out on placement creation? (See line 296 for the error)

Both the old and the new approaches are wrapped in a transaction. They aren't actually any different. We never "creat[ed] the shard row after that was considered safe"… we had an atomic transaction that resulted in other readers seeing the shard once placement rows showing up at once.

The practical reason for switching the order is because SPI enforces the foreign key constraint, so the old order is impossible (I even tried using a `DEFERRED` constraint, but I think each SPI command runs in its own subtransaction, so that failed as well).

FWIW I tested failures at each point of creation, both with the old and new implementation, and they exhibit identical functionality. I also tested visibility of modifications, and they were and are atomic.

create_shards.c
jasonmp85: Some of the CitusDB-inspired code uses unsigned integer types even though _all_ of PostgreSQL's built-in integer types are signed. We kept the code identical before to have it remain identical to CitusDB where possible, but this has been bothering me for some time now, so I've been changing unsigned types to signed as I touch any of the affected lines.

distribution_metadata.c
jasonmp85: SPI uses [its own memory context](http://www.postgresql.org/docs/9.4/static/spi-memory.html), which it switches into after connecting and which `SPI_finish` destroys.

If I needed to directly allocate some memory for a return value, I could use `SPI_palloc`, but if I'm _calling_ a function which uses plain-old `palloc`, it's easier to just switch back to the caller's memory context during that call.

distribution_metadata.c
jasonmp85: > The number is presumably the field location? This makes me a bit uncomfortable, since we are coupled to the way the query is defined in a different function

Yeah… in places that I specify a query and extract fields in the same function it felt fine, but here I agree it's kind of weird.

> How can we make the two consistent?

I don't quite get this. What's inconsistent?

> I'd rather the query just returns the fields of the table.

I wrote it that way at first and it was quite a bit less clear.

> Maybe we could then use the relation tupleDescriptor (instead of query), so we know what this function expects?

We aren't interacting with relations: we're interacting with queries. Using a relation's `tupleDescriptor` is conceptually wrong. I was careful to _always_ specify column lists in `INSERT` and `SELECT` statements so the logic is entirely unambiguous. Using a relation's descriptor with e.g. `SELECT *` relies on some assumptions I'd rather not make.

> Is there a downside to sticking with the PartitionColumn function to load the column and its type like we did before? (Instead of joining with partition).

Yeah: recursive use of SPI requires some boilerplate calls and involves more reasoning about memory contexts, etc. I avoided any recursive SPI calls in this pull request to keep away from all that.

Also, the code is less straightforward and self-similar that way. As of this pull request, the shard interval and shard placement paths are structurally identical: they run a query and map rows to objects. The old shard interval path was weirder.

distribution_metadata.c
jasonmp85: It's future-proof against changes to the array this way.

distribution_metadata.c
jasonmp85: > What is this check for? Does this mean if the table is in in the pgs_distribution_metadata namespace, it isn't considered distributed?

Exactly.

SPI exercises the full PostgreSQL machinery, including `pg_shard` itself. `pg_shard` starts every query by iterating over relations and determining whether they are distributed. If none are, we use the PostgreSQL planner/executor.

If we have a query like `SELECT * FROM foo`, we'll check whether `foo` is distributed. That requires making a `SELECT` query against `pgs_distribution_metadata.partition`. But then `pg_shard` kicks in and wants to find out whether `pgs_distribution_metadata` is distributed… an infinite loop.

By inserting a short-circuit base case for this recursion, we can escape the infinite loop. I see no reason to expect any of the distribution metadata tables will be distributed, so this is an easy way out of the recursion.

pg_shard--1.1.sql
```diff
@@ -41,6 +41,12 @@ CREATE SCHEMA pgs_distribution_metadata
 	CREATE SEQUENCE shard_id_sequence MINVALUE 10000 NO CYCLE
```
sumedhpathak: So I understand, in the SQL file if we detect CitusDB, we'll modify the view to insert using CitusDB's sequences (instead of these) right? I wanted to make sure we aren't potentially creating collisions here.

jasonmp85: Yeah; I'm in the process of rebasing just the schema changes from #103 on top of this branch so it will be much more clear what the behavior will be.

There obviously _could_ be collisions when upgrading from PostgreSQL + `pg_shard` to CitusDB, but that's a pretty obscure edge case, IMO.

distribution_metadata.c
sumedhpathak: Hmm, that is pretty sucky.
We definitely need a comment here explaining what is going on.

Is there a better way to model this? (e.g. an ``IsMetadataTable()`` function, which is either called by this function, or the calling function?) I don't know if that makes it any clearer. Fundamentally, this check should only be needed for the ``partition`` table, is that correct?

distribution_metadata.c
```diff
@@ -524,62 +467,68 @@ ColumnNameToColumn(Oid relationId, char *columnName)
 
 
 /*
- * LoadShardIntervalRow finds the row for the specified shard identifier in the
- * shard table and copies values from that row into the provided output params.
+ * TupleToShardPlacement populates a ShardInterval using values from a row of
```
sumedhpathak: ``TupleToShardInterval`` (instead of ``TupleToShardPlacement``)

distribution_metadata.c
jasonmp85: > We definitely need a comment here explaining what is going on. Is there a better way to model this?

I'd rather use a comment instead of a separate function. It's effectively only one line… moving it out will hurt locality, and there's no need for the call outside of here.

By requiring callers to check first, we open ourselves up to an infinite loop if the caller fails to do so.

> Fundamentally, this check should only be needed for the partition table, is that correct?

Correct. Since its rows are the top-level objects in the graphs of distribution metadata for tables, putting the short-circuit here is sufficient.

distribution_metadata.c
jasonmp85: Ack.

distribution_metadata.c
sumedhpathak: >> Fundamentally, this check should only be needed for the partition table, is that correct?
> Correct. Since its rows are the top-level objects in the graphs of distribution metadata for tables, putting the short-circuit here is sufficient.

If that's the case, should we make the check more precise by only checking if relation name = 'pgs_distribution_metadata.partition'? That makes the comment and its context clearer?

distribution_metadata.c
sumedhpathak: Could we add a short comment explaining this? (that SPI switches the memory context?)

distribution_metadata.c
sumedhpathak: It doesn't need to be recursive right? Could we not have the callers call PartitionColumn, and pass it as an argument to this function? I feel like that gives us a better separation, rather than the query joining with the partition table. The caller then explicitly knows the function needs a partition column.

> How can we make the two consistent?
I don't quite get this. What's inconsistent?

Consistent was probably the wrong word. I was thinking of making the dependency clearer. I think if we don't join as suggested above, the code as-is is fine. The alternative I was thinking of was to have some #defines in the header file, which brings the query and the indexes closer together (in code), so its easier to tie them together logically.

distribution_metadata.c
```diff
@@ -594,16 +543,11 @@ TupleToShardPlacement(HeapTuple heapTuple, TupleDesc tupleDescriptor)
 	ShardPlacement *shardPlacement = NULL;
 	bool isNull = false;
 
-	Datum idDatum = heap_getattr(heapTuple, ATTR_NUM_SHARD_PLACEMENT_ID,
```
sumedhpathak: Why get rid of the #defined id's?

distribution_metadata.c
```diff
@@ -755,65 +675,61 @@ InsertShardPlacementRow(uint64 shardPlacementId, uint64 shardId,
...
+	Datum argValues[] = {
+		Int64GetDatum(shardPlacementId),
+		Int32GetDatum((int32) newState)
+	};
+	const int argCount = sizeof(argValues) / sizeof(argValues[0]);
+	int spiStatus = 0;
+
+	SPI_connect();
+
+	spiStatus = SPI_execute_with_args("UPDATE pgs_distribution_metadata.shard_placement "
```
sumedhpathak: This is somewhat unrelated to this change, but how do we manage consistency of locking between Citus and pg_shard? If we are now accessing CitusDB's metadata, Citus acquires locks for certain metadata operations, but pg_shard doesn't seem to? (Or is update getting a lock on the relation for us?)

distribution_metadata.c
jasonmp85: Yup.

distribution_metadata.c
jasonmp85: > It doesn't need to be recursive right? Could we not have the callers call PartitionColumn, and pass it as an argument to this function? I feel like that gives us a better separation, rather than the query joining with the partition table. The caller then explicitly knows the function needs a partition column.

Sorry, by recursive, I meant calling an SPI-using function from within another SPI-enabled section. But what you're describing sounds great. I'll do that.

distribution_metadata.c
jasonmp85: Yes, I'll make a comment and check explicitly for the partition table.

distribution_metadata.c
jasonmp85: Yup.

distribution_metadata.c
jasonmp85: They're no longer tied to the table definition, but rather to the query order. As here is another place where the query lives in a separate function from the code extracting values, I see that preserving them could make sense. I'll put them back, I think…

distribution_metadata.c
jasonmp85: I don't think we've actually had a coherent strategy for when to lock around metadata modifications. That's a different issue than the one addressed by this PR, but I agree it must be addressed before the next release; fortunately, I already know what we need to do, and it's pretty simple, so I can open another PR for that at some point.

create_shards.c
jasonmp85: Not changing in this review, but I will go through and eradicate unsigned integers from `pg_shard` in another branch. (done)

create_shards.c
jasonmp85: Discussed at length over the past few weeks. Sticking with this new approach. (done)

distribution_metadata.c
jasonmp85: Put back `#define`s for queries and numbers where the query and number did not appear in the same function body.

PR #113, opened 2015-05-25
Push simple aggregates down 

onderkalaci:
This PR is intended for the following : 
Users should be able to exploit computational power of worker nodes in servicing aggregates. At minimum, if a `GROUP BY` is present which references the partition column, we should be able to create a query containing certain safe aggregates.

### DIFF
--- a/pg_shard.c
+++ b/pg_shard.c
@@ -7,7 +7,7 @@
  *
  * Copyright (c) 2014-2015, Citus Data, Inc.
  *
- ***-------------------------------------------------------------------------
+ ****-------------------------------------------------------------------------
  */
 
 #include "postgres.h"
@@ -111,7 +111,7 @@ static List * BuildDistributedTargetList(Query *query, List *localRestrictList,
 static Query * BuildLocalQuery(Query *query, List *localRestrictList,
 							   bool safeToPushDownAggregate);
 static Query * RemoveAggregates(Query *aggregatedQuery);
-static Node * NonAggregateExpressionMutator(Node *originalNode, AttrNumber *columnId);
+static Node * VarNodeMutator(Node *originalNode, AttrNumber *columnId);
 static PlannedStmt * PlanSequentialScan(Query *query, int cursorOptions,
 										ParamListInfo boundParams);
 static List * QueryRestrictList(Query *query);
@@ -119,11 +119,12 @@ static Const * ExtractPartitionValue(Query *query, Var *partitionColumn);
 static bool ExtractFromExpressionWalker(Node *node, List **qualifierList);
 static List * QueryFromList(List *rangeTableList);
 static List * TargetEntryList(List *expressionList);
-static List * VarTypedTargetList(List *targetEntryList);
+static List * TargetEntryVarList(List *targetEntryList);
 static CreateStmt * CreateTemporaryTable(Query *localQuery, Query *distributedQuery,
 										 bool pushDownAggregates);
 static CreateStmt * CreateTemporaryTableStmtFromQuery(Query *query);
 static CreateStmt * CreateTemporaryTableLikeStmt(Oid sourceRelationId);
+CreateStmt * CreateStatement(RangeVar *relation, List *columnDefinitionList);
 static DistributedPlan * BuildDistributedPlan(Query *query, List *shardIntervalList);
 
 /* executor functions forward declarations */
@@ -300,12 +301,12 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
 			/*
 			 * distributedPlanTargetList is used for determining intermediate
 			 * temporary table's tuple descriptor in the executer. So, if aggregates
-			 * pushed down, we need to update it to make localQuery to be executed
-			 * on the intermedita table.
+			 * pushed down, we need to update the list to enable localQuery to be
+			 * executed on the intermediate table.
 			 */
 			if (safeToPushDownAggregate)
 			{
-				distributedPlanTargetList = list_copy(localQuery->targetList);
+				distributedPlanTargetList = TargetEntryVarList(localQuery->targetList);
 			}
 			else
 			{
@@ -900,22 +901,22 @@ BuildDistributedTargetList(Query *query, List *localRestrictList,
 				List *targetVarList = pull_var_clause((Node *) targetExpression,
 													  aggregateBehavior,
 													  placeHolderBehavior);
-				ListCell *targetExpressionListCell = NULL;
+				ListCell *targetVarListCell = NULL;
 
-				foreach(targetExpressionListCell, targetVarList)
+				foreach(targetVarListCell, targetVarList)
 				{
-					/* this is equivalent of pull_var_clause for non agg and var values */
-					TargetEntry *aggregatedTargetEntry = NULL;
-					Var *targetVar = (Var *) lfirst(targetExpressionListCell);
+					/* this is equivalent of pulling Vars from non-aggregated expressions */
+					TargetEntry *newTargetEntry = NULL;
+					Var *targetVar = (Var *) lfirst(targetVarListCell);
 
-					aggregatedTargetEntry = makeTargetEntry((Expr *) targetVar,
-															-1,
-															targetListEntry->resname,
-															targetListEntry->resjunk);
-					aggregatedTargetEntry->ressortgroupref = targetListEntry->ressortgroupref;
-					aggregatedTargetEntry->resjunk = targetListEntry->resjunk;
+					newTargetEntry = makeTargetEntry((Expr *) targetVar, -1,
+													 targetListEntry->resname,
+													 targetListEntry->resjunk);
 
-					targetList = list_append_unique(targetList, aggregatedTargetEntry);
+					newTargetEntry->ressortgroupref = targetListEntry->ressortgroupref;
+					newTargetEntry->resjunk = targetListEntry->resjunk;
+
+					targetList = list_append_unique(targetList, newTargetEntry);
 				}
 			}
 		}
@@ -1004,9 +1005,9 @@ static Query *
 RemoveAggregates(Query *aggregatedQuery)
 {
 	Query *nonAggregatedQuery = copyObject(aggregatedQuery);
-	Index tableOrder = 1;
+	Index masterTableId = 1;
 	AttrNumber columnId = 1;
-	List *aggregatedTargetList = nonAggregatedQuery->targetList;
+	List *aggregatedTargetList = aggregatedQuery->targetList;
 	ListCell *targetListCell = NULL;
 	List *nonAggregatedTargetList = NIL;
 	AttrNumber targetResNo = 1;
@@ -1014,51 +1015,41 @@ RemoveAggregates(Query *aggregatedQuery)
 	foreach(targetListCell, aggregatedTargetList)
 	{
 		TargetEntry *targetListEntry = (TargetEntry *) lfirst(targetListCell);
-		TargetEntry *newTargetEntry = NULL;
+		TargetEntry *newTargetEntry = copyObject(targetListEntry);
 
-		if (targetListEntry->resjunk == true)
+		/* eliminate from final target list */
+		if (newTargetEntry->resjunk == true)
 		{
 			continue;
 		}
 
-		if (IsA(targetListEntry->expr, Aggref))
+		/* update target list entries which are Aggrefs */
+		if (IsA(newTargetEntry->expr, Aggref))
 		{
-			Var *targetVar = makeVarFromTargetEntry(tableOrder, targetListEntry);
+			Var *targetVar = makeVarFromTargetEntry(masterTableId, targetListEntry);
 
 			newTargetEntry = makeTargetEntry((Expr *) targetVar,
-													   targetResNo,
-													   targetListEntry->resname,
-													   targetListEntry->resjunk);
-
-			newTargetEntry->expr = (Expr *) NonAggregateExpressionMutator(
-										(Node *) newTargetEntry->expr, &columnId);
-
+											 targetResNo,
+											 targetListEntry->resname,
+											 targetListEntry->resjunk);
 		}
-		else
-		{
-			targetListEntry->expr = (Expr *) NonAggregateExpressionMutator(
-				(Node *) targetListEntry->expr, &columnId);
-			newTargetEntry = targetListEntry;
-		}
+
+		/* update column attribute numbers */
+		newTargetEntry->expr = (Expr *) VarNodeMutator((Node *) newTargetEntry->expr,
+													   &columnId);
 
 		/* ressortgroupref must be updated for ORDER BY columns */
-		if (newTargetEntry != NULL && targetListEntry->ressortgroupref != 0)
-		{
-			newTargetEntry->ressortgroupref = targetListEntry->ressortgroupref;
-		}
+		newTargetEntry->ressortgroupref = targetListEntry->ressortgroupref;
 
 		/*
 		 * Columns which appear on GROUP BY/ORDER BY statements show up in the
 		 * targetList. We discard these columns from the aggregated query's
 		 * target list.
 		 */
-		if (targetListEntry->resjunk == false)
-		{
-			newTargetEntry->resno = targetResNo;
-			nonAggregatedTargetList = lappend(nonAggregatedTargetList,
-											  newTargetEntry);
-			++targetResNo;
-		}
+		newTargetEntry->resno = targetResNo;
+		nonAggregatedTargetList = lappend(nonAggregatedTargetList,
+										  newTargetEntry);
+		++targetResNo;
 	}
 
 	nonAggregatedQuery->targetList = list_copy(nonAggregatedTargetList);
@@ -1071,12 +1062,12 @@ RemoveAggregates(Query *aggregatedQuery)
 
 
 /*
- * MasterAggregateMutator walks over the original non-aggregate expression,
- * and recurses into the expression mutator function for all other expression types
- * to update the varattno of the Vars in the expression.
+ * VarNodeMutator walks over the original expression,  and recurses into
+ * the expression mutator function for all other expression types to update the
+ * varattno of the Vars in the expression.
  */
 static Node *
-NonAggregateExpressionMutator(Node *originalNode, AttrNumber *columnId)
+VarNodeMutator(Node *originalNode, AttrNumber *columnId)
 {
 	Node *newNode = NULL;
 	if (originalNode == NULL)
@@ -1096,7 +1087,7 @@ NonAggregateExpressionMutator(Node *originalNode, AttrNumber *columnId)
 	}
 	else
 	{
-		newNode = expression_tree_mutator(originalNode, NonAggregateExpressionMutator,
+		newNode = expression_tree_mutator(originalNode, VarNodeMutator,
 										  (void *) columnId);
 	}
 
@@ -1302,40 +1293,19 @@ TargetEntryList(List *expressionList)
 
 
 /*
- * VarTypedTargetList get a target list and creates a new target list entry
+ * TargetEntryVarList get a target list and creates a new target list entry
  * whose elements are Var corresponding elements of the input target list.
  */
 static List *
-VarTypedTargetList(List *targetEntryList)
+TargetEntryVarList(List *targetEntryList)
 {
-	//return list_copy(targetEntryList);
-	List *smotthedTargetEntryList = NIL;
-	ListCell *targetCell = NULL;
-
 	PVCAggregateBehavior aggregateBehavior = PVC_RECURSE_AGGREGATES;
 	PVCPlaceHolderBehavior placeHolderBehavior = PVC_RECURSE_PLACEHOLDERS;
 
-	/* as well as any used in projections (GROUP BY, etc.) */
 	List *projectColumnList = pull_var_clause((Node *) targetEntryList,
 											  aggregateBehavior, placeHolderBehavior);
 
 	return TargetEntryList(projectColumnList);
-
-	foreach(targetCell, targetEntryList)
-	{
-		TargetEntry *targetEntry = lfirst(targetCell);
-
-		Var *targetVar = makeVarFromTargetEntry(1, targetEntry);
-		TargetEntry *updatedTargetEntry = makeTargetEntry((Expr *) targetVar,
-														  targetEntry->resno,
-														  targetEntry->resname,
-														  targetEntry->resjunk);
-
-		smotthedTargetEntryList = lappend(smotthedTargetEntryList, updatedTargetEntry);
-	}
-
-
-	return smotthedTargetEntryList;
 }
 
 
@@ -1364,8 +1334,7 @@ CreateTemporaryTable(Query *localQuery, Query *distributedQuery, bool pushDownAg
 
 /*
  * CreateTemporaryTableStmtFromQuery returns a CreateStmt node which will create
- * a temporary table whose columns are the same as the query parameter's range table
- * entries.
+ * a temporary table whose columns are equivalent to the target list.
  */
 static CreateStmt *
 CreateTemporaryTableStmtFromQuery(Query *query)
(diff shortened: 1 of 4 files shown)


pg_shard.c
```diff
@@ -7,7 +7,7 @@
  *
  * Copyright (c) 2014-2015, Citus Data, Inc.
  *
- *-------------------------------------------------------------------------
+ *****-------------------------------------------------------------------------
```
jasonmp85: ???

pg_shard.c
```diff
@@ -229,9 +249,11 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
 	{
 		DistributedPlan *distributedPlan = NULL;
 		Query *distributedQuery = copyObject(query);
+		List *distributedPlanTargetList = distributedQuery->targetList;
 		List *queryShardList = NIL;
 		bool selectFromMultipleShards = false;
 		CreateStmt *createTemporaryTableStmt = NULL;
+		Oid distributedTableId = InvalidOid;
```
jasonmp85: Why move this declaration? I don't see where it's needed in this outer scope.

pg_shard.c
```diff
@@ -258,21 +280,42 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
 		selectFromMultipleShards = SelectFromMultipleShards(query, queryShardList);
 		if (selectFromMultipleShards)
 		{
-			Oid distributedTableId = InvalidOid;
```
jasonmp85: Just assign the actual value here if we're going to need it in this scope (and I think we will?)

pg_shard.c
```diff
@@ -258,21 +280,42 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
...
+			 * distributedPlanTargetList is used for determining intermediate
+			 * temporary table's tuple descriptor in the executer. So, if aggregates
+			 * pushed down, we need to update the list to enable localQuery to be
+			 * executed on the intermediate table.
+			 */
+			if (safeToPushDownAggregate)
+			{
+				distributedPlanTargetList = TargetEntryVarList(localQuery->targetList);
+			}
+			else
```
jasonmp85: This `else` block feels extraneous; after all, we've already assigned the target list in the outer scope.

pg_shard.c
```diff
@@ -108,7 +123,12 @@ static Const * ExtractPartitionValue(Query *query, Var *partitionColumn);
 static bool ExtractFromExpressionWalker(Node *node, List **qualifierList);
 static List * QueryFromList(List *rangeTableList);
 static List * TargetEntryList(List *expressionList);
+static List * TargetEntryVarList(List *targetEntryList);
+static CreateStmt * CreateTemporaryTable(Query *localQuery, Query *distributedQuery,
```
jasonmp85: This function name is misleading; it doesn't create a table, it creates a create table statement.

Maybe `CreateTemporaryTableStmt`?

pg_shard.c
```diff
@@ -329,6 +373,73 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
...
+SafeToPushDownAggregate(Query *query, Oid distributedTableId)
+{
+	List *groupClauseList = query->groupClause;
+	List *targetList = query->targetList;
+	bool safeToPushDownAggregates = false;
+	Var *partitionColumn = PartitionColumn(distributedTableId);
+	char *finalAttributeName = get_attname(distributedTableId, list_length(targetList));
+
+	/*
+	 * standard_planner will be called on localQuery at PlanSequentialScan function.
```
jasonmp85: This feels like a pretty bizarre restriction to me. Why is it required? When is it even triggered?

pg_shard.c
```diff
@@ -329,6 +373,73 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
...
+ * of this function.
+ */
+static bool
+SafeToPushDownAggregate(Query *query, Oid distributedTableId)
+{
+	List *groupClauseList = query->groupClause;
+	List *targetList = query->targetList;
+	bool safeToPushDownAggregates = false;
+	Var *partitionColumn = PartitionColumn(distributedTableId);
+	char *finalAttributeName = get_attname(distributedTableId, list_length(targetList));
```
jasonmp85: This method of detection is weird. Can you just get the number of attributes in the table and use `(list_length(target_list) > numOfTableAttrs)` or something like that?

pg_shard.c
```diff
@@ -329,6 +373,73 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
...
 /*
+ * SafeToPushDownAggregate returns true if GROUP BY clause includes the partition column.
+ * This function returns false if the total length of target list is greater than the
+ * total number of columns on the distributed table whose Oid is the second parameter
+ * of this function.
+ */
+static bool
+SafeToPushDownAggregate(Query *query, Oid distributedTableId)
+{
+	List *groupClauseList = query->groupClause;
```
jasonmp85: Many of these variables can have their scopes reduced (put them in the `else` block).

pg_shard.c
```diff
@@ -329,6 +373,73 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
...
+	bool groupedByColumn = false;
+	ListCell *groupClauseCell = NULL;
+
+	foreach(groupClauseCell, groupClauseList)
+	{
+		SortGroupClause *groupClause = (SortGroupClause *) lfirst(groupClauseCell);
+		TargetEntry *groupTargetEntry = get_sortgroupclause_tle(groupClause, targetList);
+
+		Expr *groupExpression = (Expr *) groupTargetEntry->expr;
+		if (IsA(groupExpression, Var))
```
jasonmp85: Under what circumstances is this variable _not_ a `Var`? Maybe add comments to say what the possibilities are?

pg_shard.c
```diff
@@ -807,11 +1081,135 @@ BuildLocalQuery(Query *query, List *localRestrictList)
...
+	nonAggregatedQuery->hasAggs = false;
+	nonAggregatedQuery->groupClause = NIL;
+	nonAggregatedQuery->havingQual = NULL;
+
+	return nonAggregatedQuery;
+}
+
+
+/*
+ * VarNodeMutator walks over the original expression,  and recurses into
```
jasonmp85: This description is so generic I couldn't understand what this function actually does.

pg_shard.c
```diff
@@ -807,11 +1081,135 @@ BuildLocalQuery(Query *query, List *localRestrictList)
...
+		return NULL;
+	}
+
+	if (IsA(originalNode, Var))
+	{
+		uint32 masterTableId = 1; /* one table on the master node */
+		Var *newColumn = copyObject(originalNode);
+		newColumn->varno = masterTableId;
+		newColumn->varattno = (*columnId);
+		(*columnId)++;
```
jasonmp85: I realize this construct is similar to some CitusDB code, but there is no comment explaning what the `columnId` actuall is. It's some sort of state variable that lets us count during iteration of the walker? Again, I'm not really clear what's happening here, or why.

pg_shard.c
```diff
@@ -1004,6 +1402,135 @@ TargetEntryList(List *expressionList)
 
 
 /*
+ * TargetEntryVarList get a target list and creates a new target list entry
+ * whose elements are Vars, which are pulled from the input target list.
+ */
+static List *
+TargetEntryVarList(List *targetEntryList)
```
jasonmp85: This code is very similar to `flatten_tlist`, provided by PostgreSQL. Is there any way we could use that instead?

jasonmp85:
I added my first round of questions and comments here. I also have some high-level questions:

  * How much of this comes (mostly unmodified) from CitusDB? I kept stumbling across pieces that were also in CitusDB somewhere, so it was unclear what is yours and what is not. It might help my code review to know which pieces are entirely novel and which are reused.
  * The first function I came across with the `safeToPushDownAggregates` flag made me think of the rule that boolean flags on methods is an antipattern. And so I was already thinking this when I came across many more methods that do the same thing. Is there _any_ thing you can do to get rid of the boolean flag that we're pushing all the way down the stack? Passing in a boolean flag is basically the caller saying "I know you have a conditional somewhere in your implementation, please take this branch". Might as well hoist up the conditional and have two methods, where appropriate. This might be a Code Complete 2 rule as well (@ozgune)?
  * Maybe instead of a flag, you classify the `SortGroupClause` list into those that should run locally and those that should run remotely? This seems kind of pedantic, and I haven't fully thought it through but I think it could help the code flow be less bifurcated. And it would be analogous to what we see with the restrict clauses.
  * How are `ORDER BY` clauses affected by this change? Would the idea of splitting `SortGroupClause` lists into remote and local versions make more sense when we think about `ORDER BY` in addition to `GROUP BY`? Just brainstorming, I'm not actually sure.
  * Do you have any thoughts about how to start splitting `pg_shard.c` up? If we leave this as-is, the file is now around 1.5 kloc. I wasn't sure what functions were drawing from what CitusDB files, but maybe we want to mimic that structure in some places, PostgreSQL in others? At any rate, one huge planner/executor/hooks/utility file is getting burdensome.

Let's start some conversations around these questions (and maybe you can start investigating cleaning up the boolean flag issue) and I'll add more comments in the next round.

pg_shard.c
onderkalaci: @jasonmp85 Else is required since first assignment is done before BuildDistributedQuery call. Also, first assignment is required for single shard selects.

pg_shard.c
onderkalaci: I think uncrustify adds that. Fixed.

pg_shard.c
onderkalaci: Done.

pg_shard.c
onderkalaci: For example, this [test](https://github.com/citusdata/pg_shard/pull/113/files#diff-bea7cbd1e011d28d841577383c5dc2cdR264) is an example where articles table has 4 columns.

pg_shard.c
onderkalaci: Makes sense, done.

pg_shard.c
onderkalaci: Yes, I missed that.
Done.

pg_shard.c
onderkalaci: This function is directly taken from CitusDB. So, I didn't want to change it.
Should I add comment? Well, I don't know exactly why there is such a control.

pg_shard.c
onderkalaci: Yes, thats exactly the same function. Replaced.

pg_shard.c
onderkalaci: I don't want to explain it here.
I updated the function name and comment, let's see if it is easier to understand now. If still not, I can explain what the function does in detail, and we can decide on the name and comments .

onderkalaci:
@jasonmp85 

* There are two functions that come from CitusDB:
  * `GroupedByColumn` : This is exactly same with CitusDB implementation.
  * `AttributeNumberMutator` : This is inspired from `MasterAggregateMutator`

* As far as I understand from CitusDB codes, it pushes down GROUP BYs in any case (ie: does not need to be on partition column) and executes one more GROUP BY on the master table too.  So, this implementation couldn't have much in common with those codes. 

* For `safeToPushDownAggregates` flag and passing it to functions must be replaced. I totally agree with you. Code Complete 2 also states that (Section 7.2). It is mentioned as `Logical Cohesion` and considered as unacceptable. So, for now I tried to handle smaller items in your first review and concentrate on it next because I need to think about it first. If you have any suggestions, it would like to discuss them too.

* For this implementation, `ORDER BY`s never pushed down to the workers. We apply them on the local query as we do for other types of queries. Pushing down it to the workers just not work, right? But, we pull required sort columns even if it isn't on the target list. An example is on the [test](https://github.com/citusdata/pg_shard/pull/113/files#diff-1a94c07e651ce9f6517b7853ab1ebf3dR310)

onderkalaci:
@jasonmp85 I did some changes on the code flow. In the beginning, I wrote  a function whose name is `ClassifyTargetList`. That function, similar to`ClassifyRestrictions`, could classify the target lists. But, then for other parts of the code (ie: create temp table or updating groupClause of queries), I had to understand whether aggregates are pushed down or not from the target lists. That seemed as weird as previous PR. (Btw, what do you think for that approach? I can switch to it back)

Thus, in the updated PR, I wrote `ClassifyHasAggregates` functions, then, depending on that values let the code flow accordingly. So, this time, the flag does not directly shows whether we pushed down the aggregates or not. They simple show the value of `hasAgg` field of distributed and local queries.

I think this time passing those values could not be considered as anti-pattern. What do you think? 

Only `CreateTemporaryTableStmt` seems to be have an anti-pattern, but, Code complete 2 book says that if a function gets a boolean and immediately returns according to that value could not be considered as anti-pattern. So, I think  `CreateTemporaryTableStmt` seems OK from that perspective.

Also, I remembered that I also took `CreateStatement` directly from CitusDB.

pg_shard.c
jasonmp85: If it came directly from CitusDB, it's fine as-is.

jasonmp85:
I think in addition to the two you mentioned, all of these also come from CitusDB, no?

  * `GroupedByColumn`
  * `ValueToStringList`
  * `MasterTargetList`
  * `BuildCreateStatement`
  * `ColumnDefinitionList`
  * `CreateStatement`

Tell me whether these have any important changes, because from what I can see they're mostly similar to the CitusDB code and so I mostly need to look at how they're called, not what they contain.

pg_shard.c
```diff
@@ -258,21 +284,49 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
...
 			List *localRestrictList = NIL;
+			bool localQueryHasAggregates = false;
+			bool remoteQueryHasAggregates = false;
+			bool aggregatesPushedDown = false;
 
 			/* partition restrictions into remote and local lists */
 			ClassifyRestrictions(queryRestrictList, &remoteRestrictList,
 								 &localRestrictList);
 
+			/* decide remote and local hasAggs values */
```
jasonmp85: How about "decide whether remote and local queries will have aggregates".

pg_shard.c
```diff
@@ -258,21 +284,49 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
...
-			localQuery = BuildLocalQuery(query, localRestrictList);
+			distributedQuery = BuildDistributedQuery(distributedQuery,
+													 remoteRestrictList,
+													 localRestrictList,
+													 remoteQueryHasAggregates);
+			localQuery = BuildLocalQuery(query, localRestrictList,
+										 localQueryHasAggregates,
+										 remoteQueryHasAggregates);
+
+			/* determine whether we pushed down aggregates or not */
```
jasonmp85: Merge this comment into the block comment below somehow and move the assignment to be right above the `if` statement.

pg_shard.c
```diff
@@ -258,21 +284,49 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
...
+			aggregatesPushedDown = (remoteQueryHasAggregates && !localQueryHasAggregates);
+
+			/*
+			 * distributedPlanTargetList is used for determining intermediate
```
jasonmp85: How about:

> We say we are "pushing down" aggregates if only the remote query has any. The schema of the temporary table must change to reflect this, so we calculate a new target list based on the output of the pushed down aggregates in the remote query.

pg_shard.c
```diff
@@ -281,15 +335,16 @@ PgShardPlanner(Query *query, int cursorOptions, ParamListInfo boundParams)
 			 */
 			plannedStatement = PlanSequentialScan(localQuery, cursorOptions, boundParams);
 
-			/* construct a CreateStmt to clone the existing table */
-			distributedTableId = ExtractFirstDistributedTableId(distributedQuery);
-			createTemporaryTableStmt = CreateTemporaryTableLikeStmt(distributedTableId);
+			/* construct a CreateStmt depending on safeToPushDownAggregate */
```
jasonmp85: "construct a CREATE statement based on whether aggregates are being pushed down"

pg_shard.c
```diff
@@ -713,25 +768,123 @@ ClassifyRestrictions(List *queryRestrictList, List **remoteRestrictList,
 
 
 /*
- * RowAndColumnFilterQuery builds a query which contains the filter clauses from
+ * ClassifyHasAggregates decides whether local and distributed queries will include
+ * aggregates or not. remoteQueryHasAggregates and localQueryHasAggregates are output
+ * parameters to receive these boolean values.
+ */
+static void
+ClassifyHasAggregates(Query *query, bool *remoteQueryHasAggregates,
```
jasonmp85: Nothing is really being classified here. I'd say `DetermineRemoteAndLocalAggregatePresence`. Is that too long? It is much better than classify.

pg_shard.c
```diff
@@ -713,25 +768,123 @@ ClassifyRestrictions(List *queryRestrictList, List **remoteRestrictList,
...
+{
+	*remoteQueryHasAggregates = false;
+	*localQueryHasAggregates = false;
+
+	if (query->groupClause != NIL)
+	{
+		Oid tableId = ExtractFirstDistributedTableId(query);
+		bool safeToPushDownAggregate = SafeToPushDownAggregate(query, tableId);
+
+		/* if we can push down the aggregates update */
```
jasonmp85: This comment doesn't make any sense to me. I'd just remove it: the code is clear as-is.

pg_shard.c
```diff
@@ -713,25 +768,123 @@ ClassifyRestrictions(List *queryRestrictList, List **remoteRestrictList,
...
+		{
+			*remoteQueryHasAggregates = false;
+			*localQueryHasAggregates = true;
+		}
+	}
+}
+
+
+/*
+ * SafeToPushDownAggregate returns true if GROUP BY clause includes the partition column.
```
jasonmp85: How about this edit:

> `SafeToPushDownAggregate` determines whether a query's `GROUP BY` clause includes the partition column of the distributed table identified by the second argument. This function returns false if the total length of the target list is greater than the number of columns in the specified table or if the query has no `GROUP BY` clause whatsoever.

pg_shard.c
```diff
@@ -713,25 +768,123 @@ ClassifyRestrictions(List *queryRestrictList, List **remoteRestrictList,
...
+	bool safeToPushDownAggregates = false;
+	List *targetList = query->targetList;
+	List *groupClauseList = query->groupClause;
+	List *rangeTableList = query->rtable;
+	RangeTblEntry *rangeTableEntry = (RangeTblEntry *) linitial(rangeTableList);
+	int tableColumnCount = list_length(rangeTableEntry->eref->colnames);
+	int targetListCount = list_length(targetList);
+
+	/*
+	 * standard_planner will be called on localQuery at PlanSequentialScan function.
```
jasonmp85: Given the more descriptive function comment, I'd replace this with a simple _return false if there is no GROUP BY clause or it the target list has more elements than the table has columns_.

pg_shard.c
```diff
@@ -741,14 +894,170 @@ RowAndColumnFilterQuery(Query *query, List *remoteRestrictList, List *localRestr
 	fromExpr->quals = (Node *) make_ands_explicit((List *) remoteRestrictList);
 	fromExpr->fromlist = QueryFromList(rangeTableList);
 
+	if (remoteQueryHasAggregates)
+	{
+		distributedQuery->hasAggs = true;
+		distributedQuery->groupClause = list_copy(query->groupClause);
+		distributedQuery->targetList =
+			BuildAggregatedDistributedTargetList(query, localRestrictList);
```
jasonmp85: These dual line assignments are an eyesore. Any way to assign to a local variable and then assign to the field later?

pg_shard.c
```diff
@@ -741,14 +894,170 @@ RowAndColumnFilterQuery(Query *query, List *remoteRestrictList, List *localRestr
...
+	 *  Sort columns which does not appear in the final
+	 *  target list must be pulled from the worker nodes.
+	 */
+	foreach(targetListCell, newTargetList)
+	{
+		TargetEntry *targetListEntry = (TargetEntry *) lfirst(targetListCell);
+
+		if (targetListEntry->resjunk)
+		{
+			bool sortTargetEntry = SortTargetEntry(sortClause, targetListEntry);
```
jasonmp85: Change to `sortClauseContainsTargetEntry`

pg_shard.c
```diff
@@ -741,14 +894,170 @@ RowAndColumnFilterQuery(Query *query, List *remoteRestrictList, List *localRestr
...
+	return newTargetList;
+}
+
+
+/*
+ * SortTargetEntry returns true if targetEntry references one of the elements
+ * in the sortClause.
+ */
+static bool
+SortTargetEntry(List *sortClause, TargetEntry *targetEntry)
```
jasonmp85: Let's go with `SortClauseReferencedByTargetEntry`.

pg_shard.c
jasonmp85: OK, so `sortClauseReferencedByTargetEntry` is fine.

pg_shard.c
```diff
@@ -782,36 +1091,130 @@ RowAndColumnFilterQuery(Query *query, List *remoteRestrictList, List *localRestr
...
+ * aggregates if necessary. Expects queries with a single entry in their FROM list.
  */
 static Query *
-BuildLocalQuery(Query *query, List *localRestrictList)
+BuildLocalQuery(Query *query, List *localRestrictList, bool localQueryHasAggregates,
+				bool remoteQueryHasAggregates)
 {
 	Query *localQuery = copyObject(query);
 	FromExpr *joinTree = localQuery->jointree;
+	bool aggregatesPushedDown = (remoteQueryHasAggregates && !localQueryHasAggregates);
```
jasonmp85: Now that I see this here, I fail to understand why we sometimes say `aggregatesPushedDown` and sometimes say `remoteQueryHasAggregates`. In `BuildDistributedQuery` you test the latter but here you're testing the former. Given that there are only two possible permutations, what's the difference?

pg_shard.c
```diff
@@ -741,14 +894,170 @@ RowAndColumnFilterQuery(Query *query, List *remoteRestrictList, List *localRestr
...
+}
+
+
+/*
+ * BuildAggregatedDistributedTargetList returns a list of TargetEntry for the
+ * distributed query when aggregates are pushed down. This function leaves
+ * AggRefs as it is in the target list. Also, WHERE clause columns pushed down.
+ */
+static List *
+BuildAggregatedDistributedTargetList(Query *query, List *localRestrictList)
```
jasonmp85: `Aggregated` is ambiguous. Let's go with `AggregateBearing` for queries containing aggregates and `Aggregateless` for queries without them.

pg_shard.c
```diff
@@ -1004,6 +1407,215 @@ TargetEntryList(List *expressionList)
...
+	return createStmt;
+}
+
+
+/*
+ * ValueToStringList walks over the given list of string value types, converts
+ * value types to cstrings, and adds these cstrings into a new list.
+ */
+static List *
+ValueToStringList(List *valueList)
```
jasonmp85: I don't like the name of this method (or its comment), but seeing as how it comes from within CitusDB I think we must leave it.

pg_shard.c
```diff
@@ -1004,6 +1407,215 @@ TargetEntryList(List *expressionList)
...
+}
+
+
+/*
+ * MasterTargetList uses the given worker target list's expressions, and creates
+ * a target target list for the master node. This master target list keeps the
+ * temporary table's columns on the master node.
+ */
+static List *
+MasterTargetList(List *workerTargetList)
```
jasonmp85: It's a bit confusing to mix Master/Local and Worker/Remote in the same file, but again we're stuck with whatever CitusDB calls it for this method.

pg_shard.c
```diff
@@ -1045,6 +1653,23 @@ CreateTemporaryTableLikeStmt(Oid sourceRelationId)
 
 
 /*
+ * CreateUniqueTableName creates a unique table name on every invocation. It takes
+ * no parameters and returns a StringInfo.
+ */
+static StringInfo
+CreateUniqueTableName(void)
```
jasonmp85: I don't remember this function being around earlier. What is it necessary for now (and can we just use a file static variable instead of this function-scoped one?)

pg_shard.c
jasonmp85: We decided `Build[Local|Distributed]TargetListWith(out)Aggregates` is best.

pg_shard.c
jasonmp85: Turns out this was inline in another function already for the multi-shard select written by @sumedhpathak so we'll leave it as-is.

onderkalaci:
Hey @jasonmp85,

I tried to address your feedback on the changes. 

My main concern whether I fell into the same situation we had previously. Could you please consider the changes with that perspective too? This time we do not  pass the flag to many functions, but, we still we  pass the flag to some functions.  Below is your previous comments:

`The first function I came across with the safeToPushDownAggregates flag made me think of the rule that boolean flags on methods is an antipattern. And so I was already thinking this when I came across many more methods that do the same thing. Is there any thing you can do to get rid of the boolean flag that we're pushing all the way down the stack? Passing in a boolean flag is basically the caller saying "I know you have a conditional somewhere in your implementation, please take this branch". Might as well hoist up the conditional and have two methods, where appropriate.`


Also, I realized that the comment on [targetList](https://github.com/citusdata/pg_shard/blob/develop/pg_shard.h#L65) and/or its name should be changed. What do you think? Previously, the targetList of local/target queries were the same. But, it is not anymore the case. For example, should we change it to localTargetList?  Well, it is not localTargetList as well. It  basically corresponds to the temp table's schema.

PR #114, opened 2015-05-25
Switch to signed integers for all identifiers

jasonmp85:
PostgreSQL doesn't have unsigned types, so using them when storing a value in a table doesn't make sense.

### DIFF
--- a/ddl_commands.h
+++ b/ddl_commands.h
@@ -28,9 +28,9 @@
 /* function declarations to extend DDL commands with shard IDs */
 extern List * TableDDLCommandList(Oid relationId);
 extern void AppendOptionListToString(StringInfo stringBuffer, List *optionList);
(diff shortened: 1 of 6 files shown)


jasonmp85:
Assigning to @sumedhpathak, since we discussed this a bit in #110…

jasonmp85:
Reassigning to @onderkalaci because @sumedhpathak oversubscription, etc.

@onderkalaci can you look at this? It's pretty simple.

onderkalaci:
@jasonmp85 
I looked at the changes. All look fine. 
* I ran regression tests
* Checked if this changes could lead to any problems (ie: signed-unsigned comparison problems). But, all changes look safe.
* I searched for other `uint*` usage. I see no other PostgreSQL related variable. 

We can :shipit: . I let you merge whenever you want to merge.

PR #120, opened 2015-05-29
Fix repair bug

jasonmp85:
`list_union` was inappropriate here to begin with; furthermore, it needs the elements of each list to be Nodes. If they are not, we'll get an unsafe cast and (hopefully) an `ERROR`.

Do _not_ hit "merge" on this… the pull request is against `master` and so merging would create a new release. I need to bump the version number and squash the test/fix commits before merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### pg_shard v1.1.1 (May 29, 2015) ###
+
+* Fixes shard repair bug triggered when indexes present
+
 ### pg_shard v1.1.0 (March 19, 2015) ###
 
 * Speeds up INSERTs by up to 300%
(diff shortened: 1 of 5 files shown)


jasonmp85:
Travis is going to build the tip of this branch, so you won't see the actual bug exercised. I built just the failing test locally and here is the bad output:

```
SELECT master_copy_shard_placement(20, 'localhost', 5432, '127.0.0.1', 5432);
ERROR:  unrecognized node type: 1095062083
```

After the fix, this line goes away and the test works as before. I added the indexes to the test to force the DDL replay list to contain more than one command (triggering the `list_union` bug in the description).

onderkalaci:
@jasonmp85 
This change makes sense. As you already mentioned, the correct action is to `concat` the lists. Again, as you mentioned, `list_union` is not convenient since it calls`equal()` which expects the input elements to be `Node` type, where in our case elements are `char *` .

I tested the change locally. So,  we can :shipit:

PR #122, opened 2015-06-15
Disable unused parameter compiler warning

jasonmp85:
@anarazel asked for this in #121.

This warning isn't as useful as the other unused warnings (such as
unused variables) and it was getting to be more trouble than it was
worth. This removes it entirely and cuts out any extra source that
was previously needed to opt-in to unused parameters.


jasonmp85:
Pushing this up to ensure the build passes but once it does I'll merge it, since @anarazel asked for it and I agree (so that can count as a committer and reviewer for something this trivial).

Issue #93, opened 2015-03-20
Detect incomplete installations and provide useful error


jasonmp85:
Not entirely sure whether this is a feature request or a bug report, but various emails have been hitting my inbox of users either:

  * Putting `pg_shard` in `shared_preload_libraries` but forgetting to call `CREATE EXTENSION`
  * Calling `CREATE EXTENSION` but forgetting to ever `LOAD` `pg_shard` (using the above config variable or otherwise)

These causes produce a variety of symptoms, but common ones are:

  * Running a query against a distributed table but actually hitting the local table because `pg_shard`'s hooks have not been loaded (caused by lack of `LOAD`)
  * Calling a utility command and having it fail with the error `schema "pgs_distribution_metadata" does not exist` (caused by lack of `CREATE`)

This is biting enough people that some runtime guidance could help out a lot.

jasonmp85:
To protect against users accidentally hitting a local table when `pg_shard` is not loaded, we could add a trigger to those tables (this could be part of `master_create_distributed_table`'s logic) so that any queries involving them fail with a `pg_shard` related error message explaining that the table is distributed and they must `LOAD` `pg_shard` to use it.

jasonmp85:
As for forgetting `CREATE`, this is a little tougher. One of us recently hit it while trying to run `cstore_fdw`'s regression tests when `pg_shard` was also loaded. The `COPY` commands the tests were calling triggered `pg_shard`'s logic. On one hand, we _want_ the user to know about the misconfiguration, but on the other hand, putting _anything_ in the output would break other test suites.

I'm thinking we send something _only_ to the server log (possible?) to have a record that `pg_shard` is misconfigured, but otherwise just make the hook act as a no-op if the `CREATE EXTENSION` command hasn't yet been run. After all, this is probably a more rare case, since users _must_ run `CREATE EXTENSION` to get access to the functions for distributing tables.

onderkalaci:
I have one comment on forgetting `CREATE`. For that case, isn't it enough to just return at this [line](https://github.com/citusdata/pg_shard/blob/develop/pg_shard.c#L1939) ?

If user adds pg_shard to the config and not `CREATE` it, why we bother informing him? If he decides to use pg_shard, he has to `CREATE` it. This is what makes sense for me.

PR #131, opened 2015-08-11
Use CitusDB SELECT logic by default within CitusDB

### LINKED ISSUE #132: Default to CitusDB SELECT plan when within CitusDB
We originally anticipated that customers might want to use `pg_shard`'s "simpler" `SELECT` plans, even within CitusDB, leaving open the possibility (through configuration) that they could opt-in to using CitusDB's more capable (and complex) planner. We now think this behavior is unexpected to CitusDB users, so want to change `pg_shard`'s default (out-of-the-box) behavior to use CitusDB's planner unless the user opts-out.

jasonmp85:
The `use_citusdb_select_logic` configuration parameter defaults false, which is causing confusion now that pg_shard sees and understands all CitusDB metadata. This detects CitusDB at build time and defaults the configuration to true when pg_shard is built against CitusDB.

Resolves #132

### DIFF
--- a/test/sql/00-init.sql
+++ b/test/sql/00-init.sql
@@ -4,6 +4,9 @@
 
 CREATE EXTENSION pg_shard;
 
+-- ensure the test DB defaults to pg_shard select logic
+ALTER DATABASE :DBNAME SET pg_shard.use_citusdb_select_logic TO false;
+
 -- create fake fdw for use in tests
 CREATE FUNCTION fake_fdw_handler()
 RETURNS fdw_handler
--- a/include/pg_shard.h
+++ b/include/pg_shard.h
@@ -15,12 +15,20 @@
 #include "c.h"
 
 #include "access/tupdesc.h"
+#include "catalog/indexing.h"
 #include "nodes/parsenodes.h"
 #include "nodes/pg_list.h"
 #include "nodes/plannodes.h"
 #include "lib/stringinfo.h"
 #include "utils/tuplestore.h"
 
+
+/* detect when building against CitusDB */
+#ifdef DistPartitionLogicalRelidIndexId
+#define BUILT_AGAINST_CITUSDB true
+#else
+#define BUILT_AGAINST_CITUSDB false
+#endif
 
 /* prefix used for temporary tables created on the master node */
 #define TEMPORARY_TABLE_PREFIX "pg_shard_temp_table"
--- a/src/pg_shard.c
+++ b/src/pg_shard.c
@@ -197,8 +197,8 @@ _PG_init(void)
 
 	DefineCustomBoolVariable("pg_shard.use_citusdb_select_logic",
 							 "Informs pg_shard to use CitusDB's select logic", NULL,
-							 &UseCitusDBSelectLogic, false, PGC_USERSET, 0, NULL,
-							 NULL, NULL);
+							 &UseCitusDBSelectLogic, BUILT_AGAINST_CITUSDB, PGC_USERSET,
+							 0, NULL, NULL, NULL);
 
 	DefineCustomBoolVariable("pg_shard.log_distributed_statements",
 							 "Logs each statement used in a distributed plan", NULL,
--- a/test/expected/00-init.out
+++ b/test/expected/00-init.out
@@ -2,6 +2,8 @@
 -- create extension
 -- ===================================================================
 CREATE EXTENSION pg_shard;
+-- ensure the test DB defaults to pg_shard select logic
+ALTER DATABASE :DBNAME SET pg_shard.use_citusdb_select_logic TO false;
 -- create fake fdw for use in tests
 CREATE FUNCTION fake_fdw_handler()
 RETURNS fdw_handler


test/sql/00-init.sql
```diff
@@ -4,6 +4,9 @@
 
 CREATE EXTENSION pg_shard;
 
+-- ensure the test DB defaults to CitusDB select logic
```
jasonmp85: Should say _ensure the test DB defaults to not using CitusDB select logic_

sumedhpathak:
Can this change go in the .sql file itself? How is the order of these things being called? We already have logic there to determine if we are building against Citus, and take different actions based on that. Would we be able to set this variable there?

jasonmp85:
> Can this change go in the .sql file itself?

The SQL file is an installation script, so it runs once (at installation).

I _could_ [`ALTER SYSTEM`](http://www.postgresql.org/docs/9.4/static/sql-altersystem.html) from within SQL to write out a value to be used within the cluster hosting `pg_shard`, but:

  * That's not in PostgreSQL 9.3
  * It requires superuser permissions
  * It just writes out a config file; the underlying default (in the presence of no config) is the same

At any rate, custom GUCs are defined within `_PG_init`, which runs when the extension is loaded, so to change the default value of this GUC we need to detect CitusDB _there_ somehow.

> How is the order of these things being called?

I'm not 100% sure what this means, but the argument I changed here changes the underlying default value for the boolean GUC it's declaring. It runs during `_PG_init`, which runs (because we favor `shared_preload_libraries`) in the _frontend_, before any backend process starts. This severely limits what is possible within `_PG_init` (and is why in a previous review where I needed to detect CitusDB within `_PG_init` I had to detect a CitusDB-specific GUC. This time I realized there are some `#define`s I could use instead).

> We already have logic there to determine if we are building against Citus, and take different actions based on that. Would we be able to set this variable there?

That isn't if we're _building_ against CitusDB. By the time that code runs, the extension is built and has been put in place (by a `make install`). The `.sql` file is for `CREATE EXTENSION`. So at that point the library (`.so` or whatever) is built and the default value is "baked in".

Bigger picture: all the ways of changing the "default" from a higher level (configuration, `ALTER SYSTEM`, `ALTER DATABASE`, `SET`) have the fundamental limitation that—if the user removes them—the "default as defined in the C code" will be used. The "default as defined by the C code" is set by the call in `_PG_init`, so we need to do the detection there. And because we suggest people preload our library, `_PG_init` will run in a frontend, limiting our options.

If you're worried about duplication (or two ways of detecting CitusDB), I can wrap the `#ifdef` block in a function (`BuiltAgainstCitusDB`) and call it from all the places we need to detect CitusDB.

jasonmp85:
@sumedhpathak if you're OK with this approach, I can kick this over to @onderkalaci to review.

sumedhpathak:
@jasonmp85 I'm ok with the approach. Let's pass to @onderkalaci for a review, and also test it a bit thoroughly as well.

jasonmp85:
Cleaned up and ready for review by @onderkalaci.

onderkalaci:
Hey @jasonmp85,

Before reviewing this PR, I realized a bug related to the same config variable  (well, it seems it is a bug). I note it here. Do you think it is a bug? Or am I missing anything? 

I followed the steps below (on both master and this branch:):

* Compile&install pg_shard against CitusDB:
  * sudo PATH=/opt/citusdb/4.0/bin/:$PATH make install
* Connect to the master
  * /opt/citusdb/4.0/bin/psql -d  postgres
* create pg_shard
  * CREATE EXTENSION pg_shard
* See the current value of the flag:
  * show pg_shard.use_citusdb_select_logic ;
  * postgres=# show pg_shard.use_citusdb_select_logic ;
  * pg_shard.use_citusdb_select_logic 
  * -----------------------------------
  *  on
  * (1 row)

* Disconnect from the terminal
 * ctrl+c
* Reconnect and see the current value again:
  * postgres=# show pg_shard.use_citusdb_select_logic ;
  * ERROR:  unrecognized configuration parameter "pg_shard.use_citusdb_select_logic"

onderkalaci:
Hey @jasonmp85,

I reviewed and tested the changes, all looks fine. I understand from what you wrote that there is no other way to do it without compile time detection. But, maybe in another PR, we may consider to do what you suggested to make the detection be the same for all places: `If you're worried about duplication (or two ways of detecting CitusDB), I can wrap the #ifdef block in a function (BuiltAgainstCitusDB) and call it from all the places we need to detect CitusDB.`

There is only one question in my mind. Assume we follow the steps below:
* Compile pg_shard against Postgres
* Use that pg_shard binary with CitusDB
* Now, master_create_worker_shards call crashes. 

I was trying to experiment such a corner case. But, the above steps fail on the master branch as well. I think we shouldn't be taking care of this, because we shouldn't be supporting this at all. What do you think?

If you also agree with the above, I think we're ready to :shipit:

jasonmp85:
> Before reviewing this PR, I realized a bug related to the same config variable (well, it seems it is a bug). I note it here. Do you think it is a bug? Or am I missing anything?

The other day I noticed that the Bi-Directional Replication plugin raises an error when it's not in `preload_shared_libraries`. Maybe we should do something similar.

At any rate, the issue you experienced is best solved independently, probably as part of #93.

jasonmp85:
> I was trying to experiment such a corner case. But, the above steps fail on the master branch as well. I think we shouldn't be taking care of this, because we shouldn't be supporting this at all. What do you think?

Yes, we _definitely_ don't want to support building an extension in PostgreSQL and trying to run it in CitusDB. Even without this change, other parts of compilation are dependent upon the version of PostgreSQL being used for compilation (such as compiling the "right" version of `ruleutils`. The older version might work in a newer PostgreSQL version, but perhaps some of the in-memory representation of nodes will have changed between versions and could result in subtle bugs).

Also, since this pull request is against `master` (to get this into a release ASAP), I'm extremely disinclined to solve anything but the issue at hand before a merge.

Gonna merge and release `v1.2.2`.

PR #135, opened 2015-08-16
Use master's port if worker port config absent

jasonmp85:
CitusDB allows lines in `pg_worker_list.conf` to omit the port number, whereas `pg_shard` required it: confusing (or annoying) users of both.

This changes the `sscanf` format string to parse both the node name and node port as strings. If no node port is specified (or if that token starts with `#` i.e. begins a line comment), the default port is used; otherwise, we parse the node port string into an integer in precisely the same fashion used by CitusDB.

The reviewer of this request should consult the CitusDB functions `LoadWorkerNodeList` and `ParseWorkerNodeLine` in `worker_node_manager.c` to verify that the `pg_shard` implementation faithfully emulates CitusDB's implementation. Note that CitusDB uses functions from `hba.c` that are inaccessible to `pg_shard`, which is why the implementations differ.

I've removed the port number from the `pg_worker_list.conf` generated by our test launcher in order to exercise the new behavior.

### DIFF
--- a/include/connection.h
+++ b/include/connection.h
@@ -20,8 +20,9 @@
 /* maximum duration to wait for connection */
 #define CLIENT_CONNECT_TIMEOUT_SECONDS "5"
 
-/* maximum (textual) lengths of hostname */
+/* maximum (textual) lengths of hostname and port */
 #define MAX_NODE_LENGTH 255
+#define MAX_PORT_LENGTH 10
 
 /* times to attempt connection (or reconnection) */
 #define MAX_CONNECT_ATTEMPTS 2
--- a/src/create_shards.c
+++ b/src/create_shards.c
@@ -379,9 +379,10 @@ ParseWorkerNodeFile(char *workerNodeFilename)
 	List *workerNodeList = NIL;
 	char workerNodeLine[MAXPGPATH];
 	char *workerFilePath = make_absolute_path(workerNodeFilename);
+	char *workerPatternTemplate = "%%%u[^# \t]%%*[ \t]%%%u[^# \t]";
 	char workerLinePattern[1024];
-	const int WorkerNameIndex = 0;
-	const int WorkerPortIndex = 1;
+	const int workerNameIndex = 0;
+	const int workerPortIndex = 1;
 	memset(workerLinePattern, '\0', sizeof(workerLinePattern));
 
 	workerFileStream = AllocateFile(workerFilePath, PG_BINARY_R);
@@ -393,7 +394,7 @@ ParseWorkerNodeFile(char *workerNodeFilename)
 	}
 
 	/* build pattern to contain node name length limit */
-	snprintf(workerLinePattern, sizeof(workerLinePattern), "%%%us%%*[ \t]%%%us",
+	snprintf(workerLinePattern, sizeof(workerLinePattern), workerPatternTemplate,
 			 MAX_NODE_LENGTH, MAX_PORT_LENGTH);
 
 	while (fgets(workerNodeLine, sizeof(workerNodeLine), workerFileStream) != NULL)
@@ -402,6 +403,7 @@ ParseWorkerNodeFile(char *workerNodeFilename)
 		char *linePointer = NULL;
 		uint32 nodePort = PostPortNumber; /* default port number */
 		int fieldCount = 0;
+		bool lineIsInvalid = false;
 		char nodeName[MAX_NODE_LENGTH + 1];
 		char nodePortString[MAX_PORT_LENGTH + 1];
 		memset(nodeName, '\0', sizeof(nodeName));
@@ -414,7 +416,7 @@ ParseWorkerNodeFile(char *workerNodeFilename)
 								   "length of %d", MAXPGPATH)));
 		}
 
-		/* skip leading whitespace and check for # comment */
+		/* skip leading whitespace */
 		for (linePointer = workerNodeLine; *linePointer; linePointer++)
 		{
 			if (!isspace((unsigned char) *linePointer))
@@ -423,32 +425,28 @@ ParseWorkerNodeFile(char *workerNodeFilename)
 			}
 		}
 
+		/* if the entire line is whitespace or a comment, skip it */
 		if (*linePointer == '\0' || *linePointer == '#')
 		{
 			continue;
 		}
 
 		/* parse line; node name is required, but port is optional */
-		fieldCount = sscanf(workerNodeLine, workerLinePattern, nodeName, nodePortString);
+		fieldCount = sscanf(linePointer, workerLinePattern, nodeName, nodePortString);
 
 		/* adjust field count for zero based indexes */
 		fieldCount--;
 
 		/* raise error if no fields were assigned */
-		if (fieldCount < WorkerNameIndex)
+		if (fieldCount < workerNameIndex)
 		{
-			ereport(ERROR, (errcode(ERRCODE_CONFIG_FILE_ERROR),
-							errmsg("could not parse worker node line: %s",
-								   workerNodeLine),
-							errhint("Lines in the worker node file consist of a node "
-									"name and optional port separated by whitespace. "
-									"Lines starting with a '#' character are skipped.")));
+			lineIsInvalid = true;
 		}
 
 		/* no special treatment for nodeName: already parsed by sscanf */
 
 		/* if a second token was specified, convert to integer port */
-		if (fieldCount >= WorkerPortIndex && *nodePortString != '#')
+		if (fieldCount >= workerPortIndex && *nodePortString != '#')
 		{
 			char *nodePortEnd = NULL;
 
@@ -457,14 +455,19 @@ ParseWorkerNodeFile(char *workerNodeFilename)
 
 			if (errno != 0 || (*nodePortEnd) != '\0')
 			{
-				ereport(ERROR, (errcode(ERRCODE_CONFIG_FILE_ERROR),
-								errmsg("could not parse worker node line: %s",
-									   workerNodeLine),
-								errhint("Lines in the worker node file consist of a node "
-										"name and optional port separated by whitespace. "
-										"Lines starting with a '#' character are "
-										"skipped.")));
+				lineIsInvalid = true;
 			}
+		}
+
+		if (lineIsInvalid)
+		{
+			ereport(ERROR, (errcode(ERRCODE_CONFIG_FILE_ERROR),
+							errmsg("could not parse worker node line: %s",
+								   workerNodeLine),
+							errhint("Lines in the worker node file consist of a node "
+									"name and optional port separated by whitespace. "
+									"Comments begin with a '#' character and extend to "
+									"the end of their line.")));
 		}
 
 		/* allocate worker node structure and set fields */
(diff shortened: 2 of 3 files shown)


include/connection.h
```diff
@@ -20,8 +20,9 @@
 /* maximum duration to wait for connection */
 #define CLIENT_CONNECT_TIMEOUT_SECONDS "5"
 
-/* maximum (textual) lengths of hostname */
+/* maximum (textual) lengths of hostname and port */
 #define MAX_NODE_LENGTH 255
+#define MAX_PORT_LENGTH 10
```
onderkalaci: As far as I know, [maximum port]( https://www.google.com.tr/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=maximum+port+number) number is 65535. So, isn't 5 good for MAX_PORT_LENGTH ?

src/create_shards.c
```diff
@@ -378,6 +380,8 @@ ParseWorkerNodeFile(char *workerNodeFilename)
 	char workerNodeLine[MAXPGPATH];
 	char *workerFilePath = make_absolute_path(workerNodeFilename);
 	char workerLinePattern[1024];
+	const int WorkerNameIndex = 0;
+	const int WorkerPortIndex = 1;
```
onderkalaci: We generally use lowercase letters as the initial letter.  But, I think this uppercase letters come from CitusDB and we should stick to that?

src/create_shards.c
```diff
@@ -422,16 +428,43 @@ ParseWorkerNodeFile(char *workerNodeFilename)
 			continue;
 		}
 
-		/* parse out the node name and node port */
-		parsedValues = sscanf(workerNodeLine, workerLinePattern, nodeName, &nodePort);
-		if (parsedValues != 2)
+		/* parse line; node name is required, but port is optional */
+		fieldCount = sscanf(workerNodeLine, workerLinePattern, nodeName, nodePortString);
```
onderkalaci: `sscanf` returns EOF on failure. Since EOF is -1, I can see that we catch the failures below. But, should we add a comment that we already check the failure, or would it be useless?

src/create_shards.c
```diff
@@ -422,16 +428,43 @@ ParseWorkerNodeFile(char *workerNodeFilename)
...
+		/* raise error if no fields were assigned */
+		if (fieldCount < WorkerNameIndex)
 		{
 			ereport(ERROR, (errcode(ERRCODE_CONFIG_FILE_ERROR),
 							errmsg("could not parse worker node line: %s",
 								   workerNodeLine),
 							errhint("Lines in the worker node file consist of a node "
-									"name and port separated by whitespace. Lines that "
-									"start with a '#' character are skipped.")));
+									"name and optional port separated by whitespace. "
```
onderkalaci: For the `ereport`s, I have two comments considering both errors are exactly the same:

* Can't we use a boolean variable to indicate whether there is an error or not? Then, use that variable and write the `ereport` only once?
* If that is not possible, its better that we make the layout exactly the same. Second `ereport`'s `errhint` is splitted into 4 lines, whereas the first one is splitted into 3 lines.

onderkalaci:
Hey @jasonmp85, 

I added some comments to the code. But, there are some more points that I should mention.
* Our aim is to emulate CitusDB's implementation. 
* With this change, CitusDB and pg_shard behaves *almost* the same. I observed a different behavior only on the following example: 
 * localhost 9700#test
* In the above example, there is no white-space between the port number and `#` character. CitusDB can handle that, but, pg_shard errors out. I couldn't decide where is the best place to tackle this in the code.

There are some other points, that are not relevant with this change. I write them here for only documentation purposes and awareness.
* If the same host-port in the worker list, CitusDB uses only one of them. But, pg_shard uses both lines and behaves as if there are two different workers. 
* Since both CitusDB and pg_shard uses `strtoul` function, both read the following lines with no errors (but, different port number thatn 9700), since `strtoul` can parse negative and hex numbers:
 * localhost 0x9700
 * localhost -9700

src/create_shards.c
jasonmp85: I think the existing error message is nice enough for the (hopefully mostly impossible) case of EOF here. I'd prefer to leave it as it is.

src/create_shards.c
jasonmp85: > * Can't we use a boolean variable to indicate whether there is an error or not? Then, use that variable and write the `ereport` only once?

Hm… I had considered a `goto` and decided that was too unconventional, especially since the repetition here is only once (if I had three copies of the error I'd definitely do something to cut back on repetition). But I think the boolean could work…

> * If that is not possible, its better that we make the layout exactly the same. Second `ereport`'s `errhint` is splitted into 4 lines, whereas the first one is splitted into 3 lines.

Eh, that's more because I hit the 90 character limit, but I do remember thinking it was unsightly. I'll try out the boolean.

include/connection.h
jasonmp85: That is the maximum port, but the old code accepted ten digits, I think because that's the largest integer. We're just reading the file, not validating the semantics of the values; another piece of the software will error if the port number is illegal.

src/create_shards.c
jasonmp85: Good catch. Yeah, I copy-pasted it. I was going to answer "well, these are `const`, so they are capitalized", but then I realized no where else in `pg_shard` do we capitalize `const` variables like this. Even though it's copied from Citus, this function is essentially a reimplementation (or emulation) of what Citus does.

In short, I'm going to make them lowercase and I agree.

jasonmp85:
> * Our aim is to emulate CitusDB's implementation. 
> * With this change, CitusDB and pg_shard behaves *almost* the same. I observed a different behavior only on the following example: `localhost 9700#test`
> * In the above example, there is no white-space between the port number and `#` character. CitusDB can handle that, but, pg_shard errors out. I couldn't decide where is the best place to tackle this in the code.

Bah! Good catch. What do you think about setting the first `#` character to `'\0'`? Then that part of the string is ignored.

> * If the same host-port in the worker list, CitusDB uses only one of them. But, pg_shard uses both lines and behaves as if there are two different workers.

That's probably going to remain the same for the time being, but I'll ask @sumedhpathak and @ozgune for a quick opinion on it. If I fix it, it will be in a separate code review.

> * Since both CitusDB and pg_shard uses `strtoul` function, both read the following lines with no errors (but, different port number thatn 9700), since `strtoul` can parse negative and hex numbers.

Actually, I _wanted_ to use `pg_atoi` and check for a negative return value to reject them. But that's incompatible with CitusDB. It's crazy that CitusDB passes `0` as the base, and I _have_ to think that's some sort of oversight, but again I'll ask the local historians why that is the way it is. As far as negative numbers go, `stroul` is insane. It's basically just `(unsigned long) strtol`, so passing `-5` gets you some insanely huge number and confuses the user. `pg_atoi` uses `strtol(…, 10)` and has the same error handling as us, so all we'd need to do is check for a negative value if we wish. I'll ask people about this, too.

jasonmp85:
> Good catch. What do you think about setting the first # character to '\0'? Then that part of the string is ignored.

Based on the fact that this proposal would modify the input string (so comments wouldn't appear when the error message spits out the offending line), I'm going to go another direction. I think I can do this directly with `sscanf`.

jasonmp85:
OK, so what I went with was using `sscanf` to accept a series of characters for the hostname and port _so long as they are not `#`, tab, or space_. This looks pretty familiar if you know your regexes: `[^# \t]`. Quoth the manual:

<blockquote><pre><code><strong>[</strong>     Matches a nonempty sequence of characters from the specified set of
      accepted characters; the next pointer must be a pointer to <em>char</em>, and
      there must be enough room for all the characters in the string, plus a
      terminating NUL character.  The usual skip of leading white space is
      suppressed.  The string is to be made up of characters in (or not in)
      a particular set; the set is defined by the characters between the
      open bracket [ character and a close bracket ] character.  The set
      <em>excludes</em> those characters if the first character after the open
      bracket is a circumflex <strong>^</strong>.  To include a close bracket in set, [...]
</code></pre></blockquote>

Since this mode doesn't skip leading whitespace, I changed the argument to `sscanf` from `workerNodeLine` (the whole line) to `linePointer` (which has advanced to the first non-whitespace character).

So this fixes the _`#` character not offset by spaces` problem and gets us pretty close to parity with CitusDB's parsing (which, being based on `hba.c` is too complicated to replicate here).

jasonmp85:
I decided to fix the remaining parse bugs. Now we reject negative numbers and numbers expressed in anything except decimal. In addition, I discovered `fgets` leaves trailing newlines in its result, so we need to remove them (similar to [what PostgreSQL does in `hba.c`](https://github.com/postgres/postgres/blob/REL9_4_4/src/backend/libpq/hba.c#L413-L416)).

src/create_shards.c
```diff
@@ -389,26 +394,38 @@ ParseWorkerNodeFile(char *workerNodeFilename)
...
 		{
 			ereport(ERROR, (errcode(ERRCODE_CONFIG_FILE_ERROR),
 							errmsg("worker node list file line exceeds the maximum "
 								   "length of %d", MAXPGPATH)));
 		}
 
-		/* skip leading whitespace and check for # comment */
+		/* trim trailing newlines preserved by fgets, if any */
+		linePointer = workerNodeLine + workerLineLength - 1;
+		while (linePointer >= workerNodeLine &&
```
onderkalaci: @jasonmp85 Can we get rid of this while loop if we update the `workerPatternTemplate` as follows (ie: add `\n\r` for host and port patterns?):
```C
char *workerPatternTemplate = "%%%u[^# \t\n\r]%%*[ \t]%%%u[^# \t\n\r]";
```
Does that make sense? Or, should we leave as it is?

src/create_shards.c
```diff
@@ -417,21 +434,49 @@ ParseWorkerNodeFile(char *workerNodeFilename)
...
+		/* raise error if no fields were assigned */
+		if (fieldCount < workerNameIndex)
+		{
+			lineIsInvalid = true;
+		}
+
+		/* no special treatment for nodeName: already parsed by sscanf */
+
+		/* if a second token was specified, convert to integer port */
+		if (fieldCount >= workerPortIndex && *nodePortString != '#')
```
onderkalaci: Can the following be true with the latest changes?  *nodePortString == '#'
I think it is not. If so, should we remove that check?

onderkalaci:
Hey @jasonmp85, 

I think the changes are OK now.  I only added two minor comments. Feel free to ignore those and merge the change if you think they are irrelevant. But, if you change the code, let me test once more before the merge, since I already have plenty of corner case tests.

src/create_shards.c
jasonmp85: Good catch. Removing.

src/create_shards.c
jasonmp85: I initially _did_ do this, but changed it. When we print the line as part of the error message, it also printed the newline, which made a blank line in the final output, which I found annoying. By actually removing newlines from the end of `workerNodeLine` we can print it without having `\n\n` appear (PostgreSQL adds a line on its own, so ours is redundant).

jasonmp85:
> if you change the code, let me test once more before the merge, since I already have plenty of corner case tests

I removed the `*nodePortString != '#'` check but will merge regardless, since you're right that it's clear from inspection that `nodePortString` can _never_ begin with a `#`. The `sscanf` format we provide uses exclusion to ensure `#` is never in that field, so if the field was assigned, it cannot include (or start with) a `#`. Going to merge after the CI tests pass.

PR #136, opened 2015-08-17
Name parameters in the README usage examples

jasonmp85:
Some of our customers have mistakenly created clusters where they swap their intended shard count and replication factor. PostgreSQL lets name parameters in function calls, so might as well make use of that in our documentation for clarity's sake.


jasonmp85:
Talked this over with @samay-sharma and @sumedhpathak. Will merge and consider making similar changes in official CitusDB documentation.

Issue #1, opened 2014-12-04
Removing existing shards should be easier


jberkus:
We need a function for dropping all shards of a particular table.  Otherwise the process of dismantling a sharded table in order to recreate it can be quite complex.

samay-sharma:
Just as a note, this also came up in our conversations with Neustar.

ozgune:
This issue was moved to citusdata/citus#119

PR #139, opened 2015-09-02
Prevent dropping distributed tables

onderkalaci:
This PR aims to prevent attempts to DROP distributed tables. This prevention will be circumvented if the user includes the CASCADE option in their DROP command, in which case partition row belonging to the distributed table is deleted, a notice is printed and the DROP is allowed to proceed.

Note that only *partition* metadata is removed. Shard and shard placement metadata are preserved for:
*  Better compliance with CitusDB 
* Enable users to find shards on the workers even if the distributed tables are dropped

Next step for removing/dropping is already discussed [here](https://github.com/citusdata/pg_shard/issues/1).


jasonmp85:
Closing this so it won't be hanging around indefinitely. I'll shoot you an email about it.

PR #144, opened 2015-09-09
Add more warnings to build

jasonmp85:
I had an _old_ TODO item to look into other possibly-useful compiler warnings to add to the project. This adds two such warning flags, one of which is fairly broad: [`-Wshadow`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wshadow-470) and [`-Wconversion`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wconversion-507).

`-Wshadow` warns if one variable shadows one with a greater scope, and `-Wconversion` warns for all number of potentially unintended implicit conversions (between 32- and 64-bith precision, between real and integer, between signed and unsigned, truncation, etc.)

Note: do _not_ merge this, I've opened it against a non-`develop` branch in order to get a cleaner diff (the hash token code is what finally pushed me to look into conversion warnings).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,11 @@ env:
   matrix:
     - PGVERSION=9.3
     - PGVERSION=9.4
+    - PGVERSION=9.5
     - PGVERSION=40.0
 before_install:
-  - sudo apt-get update -qq
-  - sudo update-alternatives --remove-all postmaster.1.gz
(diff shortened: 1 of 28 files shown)


jasonmp85:
Closing since the base PR is now closed.

PR #145, opened 2015-09-18
Fix and improve hash token distribution algorithm

jasonmp85:
There are `2^32` distinct "hash tokens" in our hash space, but we were using `INT32_MAX` (`2^32 - 1`) in the code instead. Because of this, shard counts which one might expect to divide evenly into the space (such as 16, 32, or 256) had fewer tokens than they should have. The remainder of the tokens were stuffed into the last shard, causing uneven load.

Though fixing the `INT32_MAX` bug solves the above case, it still doesn't deal with the remainder, which can be as large as `shardCount - 1`. We could continue stuffing it into the top shard, but I find it nicer to have all shard sizes be within one token of one another.

We previously divided the shard count into the hash token count to get a "hash token increment" and added that increment each iteration: this gives something like `shardIndex * (hashCount / shardCount)`. By changing the grouping to `(shardIndex * hashCount) / shardCount`, the issue with distributing the remainder goes away entirely and we get "nice" shards.

### DIFF
--- a/src/create_shards.c
+++ b/src/create_shards.c
@@ -168,12 +168,10 @@ master_create_worker_shards(PG_FUNCTION_ARGS)
 	char relationKind = get_rel_relkind(distributedTableId);
 	char *tableName = text_to_cstring(tableNameText);
 	char shardStorageType = '\0';
-	int32 shardIndex = 0;
 	List *workerNodeList = NIL;
 	List *ddlCommandList = NIL;
 	int32 workerNodeCount = 0;
 	uint32 placementAttemptCount = 0;
-	uint32 hashTokenIncrement = 0;
 	List *existingShardList = NIL;
 
 	/* make sure table is hash partitioned */
@@ -204,9 +202,6 @@ master_create_worker_shards(PG_FUNCTION_ARGS)
 		ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
 						errmsg("replication_factor must be positive")));
 	}
-
-	/* calculate the split of the hash space */
-	hashTokenIncrement = UINT_MAX / shardCount;
 
 	/* load and sort the worker node list for deterministic placement */
 	workerNodeList = ParseWorkerNodeFile(WORKER_LIST_FILENAME);
@@ -245,7 +240,7 @@ master_create_worker_shards(PG_FUNCTION_ARGS)
 		shardStorageType = SHARD_STORAGE_TABLE;
 	}
 
-	for (shardIndex = 0; shardIndex < shardCount; shardIndex++)
+	for (int64 shardIndex = 0; shardIndex < shardCount; shardIndex++)
 	{
 		List *extendedDDLCommands = NIL;
 		int64 shardId = -1;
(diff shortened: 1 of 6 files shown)


test/expected/05-create_shards.out
```diff
@@ -227,27 +238,71 @@ WARNING:  could not create shard on "adeadhost:5432"
...
+WARNING:  Connection failed to adeadhost:5432
+WARNING:  could not create shard on "adeadhost:5432"
+ master_create_worker_shards 
+-----------------------------
+ 
+(1 row)
+
+\set VERBOSITY default
+-- pg_shard ensures all shards are roughly the same size
+SELECT max_value::integer-min_value::integer AS shard_size
```
jasonmp85: Compare the output of this query to that in #146 to see the difference in the algorithms.

src/create_shards.c
```diff
@@ -245,7 +240,7 @@ master_create_worker_shards(PG_FUNCTION_ARGS)
 		shardStorageType = SHARD_STORAGE_TABLE;
 	}
 
-	for (shardIndex = 0; shardIndex < shardCount; shardIndex++)
+	for (int64 shardIndex = 0; shardIndex < shardCount; shardIndex++)
```
onderkalaci: We generally don't declare variable on the loop itself. Should we move the declaration above?

test/expected/05-create_shards.out
```diff
@@ -227,27 +238,71 @@ WARNING:  could not create shard on "adeadhost:5432"
...
+	FROM pgs_distribution_metadata.shard
+	WHERE relation_id='weird_shard_count'::regclass
```
onderkalaci: Should we add white space?
`relation_id='weird_shard_count'::regclass` => `relation_id = 'weird_shard_count'::regclass`

test/expected/05-create_shards.out
```diff
@@ -227,27 +238,71 @@ WARNING:  could not create shard on "adeadhost:5432"
...
+	ORDER BY min_value::integer ASC;
+ shard_size 
+------------
```
onderkalaci: I guess you explicitly wanted to display each shard's range separately, so we don't have GROUP BY. 
Also, should we also add shard_id in the output? Just for readability purposes. You can overlook this comment if doesn't make sense.

test/expected/05-create_shards.out
onderkalaci: should we add white space: `max_value::integer-min_value::integer AS shard_size` =>
  `max_value::integer - min_value::integer AS shard_size`

src/create_shards.c
jasonmp85: I've been declaring loop variables in loops for most of my code. They're loop variables and are bound to the loop (which is the only place they have any meaning). If it were an `if` statement, we'd push their definition inside the braces: the only reason we don't for loops is because others don't consider it because it's C99.

I like it, and `cppcheck` gripes at me if I don't do it (since scope of the variable could be reduced).

jasonmp85:
Closing in favor of #146.

PR #146, opened 2015-09-18
Fix hash token distribution algorithm

jasonmp85:
There are `2^32` distinct "hash tokens" in our hash space, but we were using `INT32_MAX (2^32 - 1)` in the code instead. Because of this, shard counts which one might expect to divide evenly into the space (such as 16, 32, or 256) had fewer tokens than they should have. The remainder of the tokens were stuffed into the last shard, causing uneven load.

This fixes that by defining a 64-bit constant equal to `2^32` and then using that to calculate the distance between shards in the hash space.

Unlike #145, in "weird" shard counts, the first `n-1` shards will have one size and the final shard will have _all_ leftover tokens. This means the final shard could contain up to `n-1` more hash tokens than the other shards, which might have load implications. In #145, all shards are within one token of each other.

Look at the tests in each review to get an idea of the two approaches.

### DIFF
--- a/src/create_shards.c
+++ b/src/create_shards.c
@@ -168,12 +168,11 @@ master_create_worker_shards(PG_FUNCTION_ARGS)
 	char relationKind = get_rel_relkind(distributedTableId);
 	char *tableName = text_to_cstring(tableNameText);
 	char shardStorageType = '\0';
-	int32 shardIndex = 0;
 	List *workerNodeList = NIL;
 	List *ddlCommandList = NIL;
 	int32 workerNodeCount = 0;
 	uint32 placementAttemptCount = 0;
-	uint32 hashTokenIncrement = 0;
+	uint64 hashTokenIncrement = 0;
 	List *existingShardList = NIL;
 
 	/* make sure table is hash partitioned */
@@ -206,7 +205,7 @@ master_create_worker_shards(PG_FUNCTION_ARGS)
 	}
 
 	/* calculate the split of the hash space */
-	hashTokenIncrement = UINT_MAX / shardCount;
+	hashTokenIncrement = HASH_TOKEN_COUNT / shardCount;
 
 	/* load and sort the worker node list for deterministic placement */
 	workerNodeList = ParseWorkerNodeFile(WORKER_LIST_FILENAME);
@@ -245,7 +244,7 @@ master_create_worker_shards(PG_FUNCTION_ARGS)
 		shardStorageType = SHARD_STORAGE_TABLE;
 	}
 
-	for (shardIndex = 0; shardIndex < shardCount; shardIndex++)
+	for (int64 shardIndex = 0; shardIndex < shardCount; shardIndex++)
 	{
 		List *extendedDDLCommands = NIL;
 		int64 shardId = -1;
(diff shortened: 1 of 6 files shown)


src/create_shards.c
```diff
@@ -256,13 +255,13 @@ master_create_worker_shards(PG_FUNCTION_ARGS)
 		/* initialize the hash token space for this shard */
 		text *minHashTokenText = NULL;
 		text *maxHashTokenText = NULL;
-		int32 shardMinHashToken = INT_MIN + (shardIndex * hashTokenIncrement);
-		int32 shardMaxHashToken = shardMinHashToken + hashTokenIncrement - 1;
+		int32 shardMinHashToken = INT32_MIN + (shardIndex * hashTokenIncrement);
+		int32 shardMaxHashToken = shardMinHashToken + (hashTokenIncrement - 1);
```
jasonmp85: These parenthesis are necessary to avoid wrap, I think.

test/expected/05-create_shards.out
```diff
@@ -227,27 +238,71 @@ WARNING:  could not create shard on "adeadhost:5432"
...
+WARNING:  Connection failed to adeadhost:5432
+WARNING:  could not create shard on "adeadhost:5432"
+ master_create_worker_shards 
+-----------------------------
+ 
+(1 row)
+
+\set VERBOSITY default
+-- pg_shard ensures all shards are roughly the same size
+SELECT max_value::integer-min_value::integer AS shard_size
```
jasonmp85: Compare the output of this query to that in #145 to see the difference in the algorithms.

test/expected/05-create_shards.out
```diff
@@ -227,27 +238,71 @@ WARNING:  could not create shard on "adeadhost:5432"
...
+	FROM pgs_distribution_metadata.shard
+	WHERE relation_id='weird_shard_count'::regclass
+	ORDER BY min_value::integer ASC;
+ shard_size 
+------------
```
onderkalaci: I guess you explicitly wanted to display each shard's range separately, so we don't have GROUP BY. 
Also, should we also add shard_id in the output? Just for readability purposes. You can overlook this comment if doesn't make sense.

test/expected/05-create_shards.out
onderkalaci: should we add white space: `max_value::integer-min_value::integer AS shard_size` =>
  `max_value::integer - min_value::integer AS shard_size`

ozgune:
Quick question. Is this pull request good to go?

jasonmp85:
@ozgune I'm going to clean it up and merge it, yes. I'll get an update out to @onderkalaci for him to sign off on tonight.

test/expected/05-create_shards.out
jasonmp85: Sure.

test/expected/05-create_shards.out
jasonmp85: I _think_ shard identifiers differ between CitusDB and `pg_shard` for now, so doing so would require two separate test result files. For that reason I'll pass on adding it for now.

jasonmp85: Also, yeah, I wanted to contrast this with #145, so I showed them all. I think that's best, since it shows the full result of this algorithm.

PR #147, opened 2015-10-02
Transaction manager integration

kelvich:
We slightly changed DTM interface and now it should be called in following way:

```sql
SELECT dtm_begin_transaction(); -- Run on node1, returns int, for example 42
SELECT dtm_join_transaction(42); -- Run on node2
-- Now we can start usual transactions, snapshot will be set automatically for both nodes.
(COMMIT; COMMIT) -- commit as usual should be done in parallel
```
Also I've added PrepareDtmTransaction to distributed selects too, so now we can have consistent read transactions too.

### DIFF
--- a/src/pg_shard.c
+++ b/src/pg_shard.c
@@ -137,7 +137,7 @@ static void TupleStoreToTable(RangeVar *tableRangeVar, List *remoteTargetList,
 							  TupleDesc storeTupleDescriptor, Tuplestorestate *store);
 static void PgShardExecutorRun(QueryDesc *queryDesc, ScanDirection direction, long count);
 static int32 ExecuteDistributedModify(DistributedPlan *distributedPlan);
-static void PrepareDtmTransaction(List *taskList);
+static void PrepareDtmTransaction(Task *task);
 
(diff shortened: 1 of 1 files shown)


marcocitus:
That's a great improvement over the previous API! Does the dtm_begin_transaction need to be in the same transaction block, or does it just call the DTMD? For example, if the coordinator had the DTM extension installed, could it call dtm_begin_transaction locally?

kelvich:
dtm_begin_transaction should be called before transaction and on the node where that transaction will take place. So no, you can't call it on master but perform transaction on different node.

What we have now with global selects -- it's our banking test. When pg_shard.use_dtm_transactions is on we see only one value of total amount of money on accounts, and when pg_shard.use_dtm_transactions is off total amount frequently fluctuates during concurrent updates.

jasonmp85:
This is neat, but is it actually intended for merging, or just part of a proof-of-concept I'm not privy to?

Notably:

  * After the merge, the build fails
  * A lot of code is commented out

jasonmp85:
Oh, I completely missed that the target for this PR is _not_ `develop`. Carry on!

PR #152, opened 2015-10-26
Update REAME

onderkalaci:
This PR updates the README about increaing the INSERT throughput on pg_shard cluster.
In the last pg_shard sync meeting, we decided that mentioning the concurrent INSERTs may help users to get better performance. 

@jasonmp85 Feel free to change the wording&structure in the PR, my aim is to kick off this change.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -141,20 +141,21 @@ copy_to_distributed_table -CH -d '|' -n NULL input.csv users
 
 Call the script with the `-h` for more usage information.
 
-### Increasing INSERT throughput
+#### Maximizing Throughput
 
-To maximize INSERT throughput, you should run statements in parallel. This helps utilizing multiple CPU cores. For instance, if you want to load the contents of the `input.csv`, first split the file and then run `copy_to_distributed_table` in parallel as shown below:
+Because `INSERT` commands must wait for a response before returning, loading rows through a single connection can never fully exploit the write capacity of your `pg_shard` cluster. To achieve the highest insert rates, it is necessary to load data in a concurrent fashion. The difference, even in a small cluster, will be significant.
+
+For example, we could split an input file (`input.csv`) into chunks and load each chunk using a separate instance of `copy_to_distributed_table` (requires GNU `split`):
 
 ```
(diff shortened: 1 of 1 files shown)


doc/README.md
```diff
@@ -141,6 +141,17 @@ copy_to_distributed_table -CH -d '|' -n NULL input.csv users
 
 Call the script with the `-h` for more usage information.
 
+### Increasing INSERT throughput
+
+To maximize INSERT throughput, you should run statements in parallel. This helps utilizing multiple CPU cores. For instance, if you are loading data from two files you could run them in parallel such as the following:
+
+```
+copy_to_distributed_table -CH -d '|' -n NULL input_1.csv users &
```
marcocitus: I've gotten into the habit of just copy-pasting from https://www.citusdata.com/blog/14-marco/164-parallel-data-loading-for-pg-shard whenever I need to load data into pg_shard. Would it be worth putting that script here? It seems more practical.

onderkalaci: @marcocitus Yes, it makes sense and its much practical as you mentioned. I updated the PR. Thanks!

jasonmp85:
Ok, so I edited this for accuracy (I used _concurrent_ where we could, instead of _parallel_) and added a section about _why_ concurrent queries are a good idea (if we have to tell our users to do this, they likely don't understand the underlying cause). I also rewrote the last bit about `psql` to be about applications generally (i.e. if you have an application doing writes, figure out how many copies (threads, connections, processes, actors, whatever) of it you need).

Additionally, I removed the use of `find`, which I consider one of the worst-designed commands in Unix. `ls` does just fine there… and I removed the `echo` and `sh` in favor of `-t` and `-I%`. Finally, I mentioned that only GNU split is capable of `-n` (BSD split doesn't have that option).

But of course the build is failing because PostgreSQL 9.5 has a new beta which changes certain outputs. I'll fix that in `develop` real quick and force a rebuild here.

jasonmp85:
Gah. Small subject–verb plurality bug. Fixing, then merging.

PR #153, opened 2015-10-27
Cache SPI plans to restore performance

jasonmp85:
Version v1.2 introduced the use of SPI in the metadata system to better abstract differences between pg_shard and CitusDB representations of distribution metadata. We didn't initially use cached SPI plans because cursory tests showed limited wins given the added complexity.

Well, additional tests uncovered _severe_ performance regressions when many writers are involved. This change caches each metadata function's SPI plan in a function-local static variable on first execution, using the cached plans on subsequent executions. After this change, we can restore performance to near v1.1 levels (SPI does inevitably add _some_ overhead, no matter what).

### DIFF
--- a/src/distribution_metadata.c
+++ b/src/distribution_metadata.c
@@ -123,6 +123,7 @@ LoadShardIntervalList(Oid distributedTableId)
 	Datum argValues[] = { ObjectIdGetDatum(distributedTableId) };
 	const int argCount = sizeof(argValues) / sizeof(argValues[0]);
 	int spiStatus PG_USED_FOR_ASSERTS_ONLY = 0;
+	static SPIPlanPtr spiPlan = NULL;
 
 	/*
 	 * SPI_connect switches to its own memory context, which is destroyed by
@@ -135,8 +136,16 @@ LoadShardIntervalList(Oid distributedTableId)
 
 	SPI_connect();
 
-	spiStatus = SPI_execute_with_args(SHARD_QUERY_PREFIX " WHERE s.relation_id = $1",
-									  argCount, argTypes, argValues, NULL, false, 0);
+	if (spiPlan == NULL)
+	{
+		spiPlan = SPI_prepare(SHARD_QUERY_PREFIX " WHERE s.relation_id = $1", argCount,
+							  argTypes);
+
+		spiStatus = SPI_keepplan(spiPlan);
+		Assert(spiStatus == 0);
+	}
+
+	spiStatus = SPI_execute_plan(spiPlan, argValues, NULL, false, 0);
 	Assert(spiStatus == SPI_OK_SELECT);
 
 	oldContext = MemoryContextSwitchTo(upperContext);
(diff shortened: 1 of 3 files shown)


jasonmp85:
Oh, and please _don't_ hit the merge button here. Merging this means creating a release, and I have a local command to do merge it to master, develop, and tag and sign the release. So I'll run that after getting some eyes on this.

src/distribution_metadata.c
```diff
@@ -411,10 +452,16 @@ IsDistributedTable(Oid tableId)
...
+	if (spiPlan == NULL)
+	{
+		spiPlan = SPI_prepare("SELECT NULL FROM pgs_distribution_metadata.partition "
+							  "WHERE relation_id = $1", argCount, argTypes);
+
+		spiStatus = SPI_keepplan(spiPlan);
+		Assert(spiStatus == 0);
+	}
+
+	spiStatus = SPI_execute_plan(spiPlan, argValues, NULL, false, 1);
```
marcocitus: Here and for the other SELECTs, should read_only be null?

jasonmp85: Given that it wasn't before, I don't want to change it now. I could have had a reason for it (cargo-culting!), but I don't remember any more. Might have had something to do with undesirable snapshot semantics, but I'm honestly speculating at this point.

Were there any blockers for :shipit:?

jasonmp85: Just looked into it, and yup:

> When `read_only` is `false`, `SPI_execute` increments the command counter and computes a new _snapshot_ before executing each command in the string. The snapshot does not actually change if the current transaction isolation level is `SERIALIZABLE` or `REPEATABLE READ`, but in `READ COMMITTED` mode the snapshot update allows each command to see the results of newly committed transactions from other sessions. This is essential for consistent behavior when the commands are modifying the database.

> When `read_only` is `true`, `SPI_execute` does not update either the snapshot or the command counter, and it allows only plain `SELECT` commands to appear in the command string. The commands are executed using the snapshot previously established for the surrounding query. This execution mode is somewhat faster than the read/write mode due to eliminating per-command overhead. It also allows genuinely _stable_ functions to be built: since successive executions will all use the same snapshot, there will be no change in the results.

> It is generally unwise to mix read-only and read-write commands within a single function using SPI; that could result in very confusing behavior, since the read-only queries would not see the results of any database updates done by the read-write queries.

I wanted each command to see all committed metadata changes (the locks we have handle making sure consistent boundaries are preserved at a level more fine-grained and semantic than just a whole database snapshot), so the behavior of `read_only` is actually undesirable for us. We want these to act like a bunch of `READ COMMITTED` statements executed one-after-another.

marcocitus:
Makes sense, although the benefit seems very marginal given that we only allow single-statement transactions at the moment. :shipit:

jasonmp85:
I'm going to merge this but want to be clear about my meaning:

If a concurrent transaction adds or removes shard stuff, we want to see it _right away_. With `read_only` set to true, all the methods would use the snapshot at the beginning of the transaction, possibly resulting in stale data. With `read_only` set to false, you might read the partition column, shard list, and then have a racing transaction mark a placement as unhealthy right before you read it. `READ COMMITTED` means you'd see that unhealthiness right then, even though the commit occurred after you'd started your work (i.e. you'd already read the partition column and shard metadata). The other case, you'd still see the placement as healthy (similar to read repeatable or serializable). The locking system was designed with this immediacy and concurrency in mind, so having the data layer expose things with this isolation is best.

I guess there should be a doc about this stuff, huh? But I think we're on track to switch to caches and invalidation medium term anyhow, so ideally you'd read from a cache eventually anyways and only refresh (and not to the latest snapshot, but to what's _actually committed_ in the table at that moment) the cache upon invalidation, which has nice performance benefits anyways.

PR #155, opened 2015-12-23
Add COPY utility to pg_shard

knizhnik:


src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+#include "utils/guc.h"
+#include "utils/lsyscache.h"
+#include "utils/palloc.h"
+#include "utils/rel.h"
+#include "utils/relcache.h"
+#include "utils/snapmgr.h"
+#include "utils/tuplestore.h"
+#include "utils/memutils.h"
+
+typedef struct {
```
marcocitus: Can we modularise the transaction-manager API? Put it in a separate file and put this struct in a header file?

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+#include "utils/tuplestore.h"
+#include "utils/memutils.h"
+
+typedef struct { 
+	bool (*Begin)(PGconn* conn);
+	bool (*Prepare)(PGconn* conn, char const* relationName, int64 shardId);
+	bool (*CommitPrepared)(PGconn* conn, char const* relationName, int64 shardId);
+	bool (*RollbackPrepared)(PGconn* conn, char const* relationName, int64 shardId);
+	bool (*Rollback)(PGconn* conn);
+} PgCopyTransactionManager;
```
marcocitus: Is this specific to Copy? What about PgShardTransactionManager?

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+
+int PgCopyTMGR;
```
marcocitus: This should probably be an enum, also avoid acronyms where possible.

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+	PGresult *result;
+	bool ret = true;
+	va_list args;
+	char buf[MAX_STMT_LEN];
+
+	va_start(args, sql);
+	vsnprintf(buf, sizeof(buf), sql, args);
+	result = PQexec(conn, buf);
+
+	if (PQresultStatus(result) != expectedResult)
```
marcocitus: Should PQclear be called afterwards?

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+	return ret;
+}
+
+/*
+ * Handle copy to/from distributed table
+ */
+bool PgShardCopy(CopyStmt *copyStatement, char const* query)
+{		 
+	RangeVar *relation = copyStatement->relation;
+	ListCell *shardIntervalCell;
```
marcocitus: We initialize every variable (pointers to NULL, ints to 0, ...)

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+	ListCell *taskPlacementCell;
+	List *shardIntervalList;
+	int relationSuffixPos;
+	char *relationName;
+	char const *lowerQuery, *relationOcc;
+	PgCopyTransactionManager* tmgr = &PgCopyTmgrImpl[PgCopyTMGR];
+
+	if (relation != NULL)
```
marcocitus: To avoid deep nesting, how about we return if relation == NULL? Or do this check prior to calling PgShardCopy? It seems the code for that was already there.

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+	{
+		bool failOK = true;
+		Oid tableId = RangeVarGetRelid(relation, NoLock, failOK);
+		bool isDistributedTable = false;
+
+		isDistributedTable = IsDistributedTable(tableId);
+		if (isDistributedTable)
```
marcocitus: To avoid deep nesting, how about we return if distributedTable is false? Or do the check prior to calling PgShardCopy.

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+			Const *rightConst;
+			int64 failedShard = -1;
+			Relation rel;
+			StringInfo lineBuf;
+			CopyConnection* copyConn;
+			int i;
+
+			relationName = get_rel_name(tableId);
+			lowerQuery = lowerstr(query);
+			relationOcc =  strstr(lowerQuery, lowerstr(relationName));
```
marcocitus: This seems very fragile. How about adding a function that re-constructs a COPY string from CopyStmt?

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+				queryList = raw_parser(newQuerySubstring->data);
+				copyStatement->query = linitial(queryList);
+				copyStatement->relation = NULL;
+				/* Collecting data from shards will be done by select handler */
+				DoCopy(copyStatement, query, &processedCount);
+				return true;
+			}
+
+			/* construct pseudo predicate specifying condition for partition key */
+			partitionColumn = PartitionColumn(tableId);
```
marcocitus: This seems like boiler-plate for calling PruneShardList. How about adding a ShardsForPartitionValue(partitionValue, shardIntervalSet) function that does the "magic"? This can then be changed to use the fast join pruning logic.

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+			rightOp = get_rightop((Expr *) equalityExpr);				 
+			Assert(IsA(rightOp, Const));
+			rightConst = (Const *) rightOp;
+			rightConst->constvalue = 0;
+			rightConst->constisnull = false;
+			rightConst->constbyval = get_typbyval(columnOid);
+			whereClauseList = list_make1(equalityExpr);
+
+			/* allocate column values and nulls arrays */
+			rel = heap_open(tableId, ExclusiveLock);
```
marcocitus: Do we need an exclusive lock on the relation?

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+												 ALLOCSET_DEFAULT_MINSIZE,
+												 ALLOCSET_DEFAULT_INITSIZE,
+												 ALLOCSET_DEFAULT_MAXSIZE);
+				
+			/* 
+			 * Construct hash table used for shardId->Connection mapping.
+			 * We can not use connection cache from connection.c used by GteConnection because
+			 * we need to establish multiple connections with each nodes: one connection per shard
+			 */
+			shard2conn = CreateShard2ConnectionHash();
```
marcocitus: I guess we typically use uncool naming without numbers :), e.g. CreateShardConnectionHash 

Could you put this behind a connection hash API, maybe move the logic into a separate file.

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+			if (copyState->binary) { 
+				elog(ERROR, "Copy in binary mode is not currently supported");
+			}
+
+			PG_TRY();
+			{
+				while (nextRowFound)
+				{
+					MemoryContext oldContext = MemoryContextSwitchTo(tupleContext);
+					nextRowFound = NextCopyFrom(copyState, NULL, columnValues, columnNulls, NULL);
```
marcocitus: Is it possible to only parse the partition column by using NextCopyFromRawFields? (not sure)

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+					MemoryContextSwitchTo(oldContext);
+					
+					/* write the row to the shard */
+					if (nextRowFound)
```
marcocitus: How about exiting the loop if !nextRowFound?

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+					{		
+						if (columnNulls[partitionColumn->varattno]) {
```
marcocitus: end-of-line between ) and {

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+					MemoryContextSwitchTo(oldContext);
+					
+					/* write the row to the shard */
+					if (nextRowFound)
+					{		
+						if (columnNulls[partitionColumn->varattno]) { 
+							ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+											errmsg("cannot copy row with NULL value "
+												   "in partition column")));
+						} else {
```
marcocitus: Since the if statement always errors out, could we maybe declare the variables at the start and remove the else?

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+						if (columnNulls[partitionColumn->varattno]) { 
+							ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+											errmsg("cannot copy row with NULL value "
+												   "in partition column")));
+						} else { 
+							Datum partitionValue = columnValues[partitionColumn->varattno];
+							rightConst->constvalue = partitionValue;
+							prunedList = PruneShardList(tableId, whereClauseList, shardIntervalList);
+							
+							foreach(shardIntervalCell, prunedList)
```
marcocitus: Wouldn't this ingest duplicates?

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+							{
+								ShardInterval *shardInterval = (ShardInterval *) lfirst(shardIntervalCell);
+								int64 shardId = shardInterval->id;
+								bool found;
+								
+								copyConn = (CopyConnection*)hash_search(shard2conn, &shardInterval->id, HASH_ENTER, &found);
+								if (!found) {
```
marcocitus: This should probably be hidden behind a GetConnectionForShard function or something similar

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+								int64 shardId = shardInterval->id;
+								bool found;
+								
+								copyConn = (CopyConnection*)hash_search(shard2conn, &shardInterval->id, HASH_ENTER, &found);
+								if (!found) { 
+									List *finalizedPlacementList;
+									copyConn->nReplicas = 0;
+									copyConn->preparedMask = 0;
+									/* grab shared metadata lock to stop concurrent placement additions */
+									LockShardDistributionMetadata(shardId, ShareLock);
```
marcocitus: Shard locks should be obtained in numerical order to avoid deadlock. You can maybe reuse AcquireExecutorShardLocks.

It would help the overall structure of the code to separate it into a planning and execution phase, where the planning phase generates tasks (maybe "streams") and the execution phase executes them. This would also make it more similar to existing pg_shard code.

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+										conn = ConnectToNode(nodeName, nodePort);
+										if (conn != NULL)										 
+										{
+											copyConn->conn[copyConn->nReplicas++] = conn;
+											/*
+											 * New connection: start transaction with copy command on it.
+											 * Append shard id to table name.
+											 */
+											if (!tmgr->Begin(conn) ||
+												!PgShardExecute(conn, PGRES_COPY_IN, "%.*s_%ld%s", relationSuffixPos, query,
```
marcocitus: What if I want to COPY from a file? The user would type FROM 'filename', but we should still send FROM STDIN to the worker.

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+								lineBuf = CopyGetLineBuf(copyState);
+								lineBuf->data[lineBuf->len++] = '\n'; /* there was already new line in the buffer, but it was truncated: no need to heck available space */
+								/* Replicate row to all shards */
+								for (i = 0; i < copyConn->nReplicas; i++) { 
+									PQputCopyData(copyConn->conn[i], lineBuf->data, lineBuf->len);
+								}
+							}
+						}
+					}
+					MemoryContextReset(tupleContext);
```
marcocitus: Are there any performance drawbacks to resetting the memory context after every row?

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+				}
+				/* Perform two phase commit in replicas */
+				hash_seq_init(&hashCursor, shard2conn);
+				while (failedShard < 0 && (copyConn = (CopyConnection*)hash_seq_search(&hashCursor)) != NULL) {
```
marcocitus: Can this be separated into two sub-problems?
1) closing the COPY on all shard placements
2) preparing the transaction on all shard placements

We could also structure this more nicely if there was a task list. Something like:
    FinalizeCopyStreams(copyStreamList);

```
if (transactionManager != NULL)
{
    PrepareCopyTransaction(transactionManager, copyStreamList);
}
```

minor: Since break; is used immediately after setting failedShard, it seems like failedShard < 0 will never be false.

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+					for (i = 0; i < copyConn->nReplicas; i++) {						
+						PQputCopyEnd(copyConn->conn[i], NULL);
+						if (tmgr->Prepare(copyConn->conn[i], relationName, copyConn->shardId)) { 
+							copyConn->preparedMask |= 1 << i;
```
marcocitus: Would be nicer if this was a flag on a copy task.

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+				}
+			}
+			PG_CATCH(); /* do recovery */
+			{
+				EndCopyFrom(copyState);
+				heap_close(rel, ExclusiveLock);
+
+				/* Rollback transactions */
+				hash_seq_init(&hashCursor, shard2conn);
+				while ((copyConn = (CopyConnection*)hash_seq_search(&hashCursor)) != NULL) {
```
marcocitus: Same comments as above about code structure.

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+
+			EndCopyFrom(copyState);
+			heap_close(rel, ExclusiveLock);
+
+			/* Complete two phase commit */
+			hash_seq_init(&hashCursor, shard2conn);
+			while ((copyConn = (CopyConnection*)hash_seq_search(&hashCursor)) != NULL) {
+				for (i = 0; i < copyConn->nReplicas; i++) {						
+					if (copyConn->preparedMask & (1 << i)) { 
+						tmgr->RollbackPrepared(copyConn->conn[i], relationName, copyConn->shardId);
```
marcocitus: Uuh, always rolls back? :)

src/pg_shard.c
```diff
@@ -2104,50 +2111,8 @@ PgShardProcessUtility(Node *parsetree, const char *queryString,
...
-			bool failOK = true;
-			Oid tableId = RangeVarGetRelid(relation, NoLock, failOK);
-			bool isDistributedTable = false;
-
-			Assert(rawQuery == NULL);
-
-			isDistributedTable = IsDistributedTable(tableId);
-			if (isDistributedTable)
-			{
-				ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
```
marcocitus, line 2122: How about we enter the PgShardCopy logic here?

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+											copyConn->conn[copyConn->nReplicas++] = conn;
+											/*
+											 * New connection: start transaction with copy command on it.
+											 * Append shard id to table name.
+											 */
+											if (!tmgr->Begin(conn) ||
+												!PgShardExecute(conn, PGRES_COPY_IN, "%.*s_%ld%s", relationSuffixPos, query,
+																(long)shardId, query + relationSuffixPos))
+											{
+												continue;
```
marcocitus: Under what circumstances does this happen? Should we fail here?

src/pg_copy.c
```diff
@@ -0,0 +1,626 @@
...
+											}
+										} else { 
+											elog(ERROR, "Failed to connect to node %s:%s", nodeName, nodePort);
```
marcocitus: Nodes may be down, in that case we should mark the placement as inactive if the COPY completes.

src/pg_copy.c
knizhnik: Originally PgCopyTransactionManager was in pg_copy.h and there was a pointer to current TMGR. But then I decided to make it possible to switch TMGR using GUC variable. So I created array with different TMGR implementations and "hide" it together with structure definition in pg_copy.

src/pg_copy.c
knizhnik: Right now it is used only in COPY. But actually there is nothing COPY specific in it. As you can see - it is just encapsulation of two-pahse commit. It can be used in other pg_shard update statements.

src/pg_shard.c
knizhnik, line 2122: Frankly speaking I do not think that it is good idea: I prefer to have all logic related with handling  copy in one place. 
Moreover, splitting logic will cause some code duplication: for example extraction of tableId.  
Also, you asked me to put all new code in new file...

src/pg_copy.c
knizhnik: This code was taken from cstore_fdw. I looked at AllocSetReset implementation - it seems to be smart enough to not deallocate all blocks -  keeper block is preserved. So, if row is small enough and fit in one block, then there should be no repeated mallocs-frees.

src/pg_copy.c
knizhnik: Do I need to lock shard data or shard metadata?
AcquireExecutorShardLocks is calling method LockShardData.
And I grab code from BuildDistributedPlan which is calling  LockShardDistributionMetadata.
Also may be I missed something, but the list for DistributedQueryShardList is build by DistributedQueryShardList and I do not see any sorting by shardId here...
It does the same as copy: LookupShardIntervalList + PruneShardList.
I wonder why it is not possible to return ordered list in LookupShardIntervalList?

src/pg_copy.c
knizhnik: Sorry, do not understand how it can ingest duplicates.

src/pg_copy.c
knizhnik: Need to cleanup context

src/pg_copy.c
knizhnik: It can be done if parsing values will be bottleneck.

src/pg_copy.c
knizhnik: Sorry, in this case we will have to reconstruct where condition for each row.

src/pg_copy.c
knizhnik: Yehhh...
Can't say that I almost existed by this code. Especially after replacing file name with STDIN. 
But reconstruction query string from options seems to be also non trivial task requiring much more code. And what is worser - if syntax of COPY command will be extended, this code also has to be changed...

src/pg_copy.c
knizhnik: The question is whether somebody else except copy need to have connection cache with multiple connections to each node? If not, then it is not a good idea to keep all this connections alive after end of the COPY. 

It is possible to extract this connection pool to separate module or place in connection.c. But right now the idea is that this cache is filled during copy and is completely destructed at end of copy. And we do not need to worry about memory deallocation, because everything  us done inside statement context.

src/pg_copy.c
knizhnik: Should we completely reject COPY in this case?
Otherwise we will loose consistency: not all input  data is populated to all replicas.

src/pg_copy.c
knizhnik: Sorry, I do not understand completely your idea. What should be linked in copyStreamList?
Frankly speaking, I do not think that it can simplify the code or make it more reusable...
May be I am wrong, because I do not understand what you mean.

src/pg_copy.c
marcocitus: What if the schema name contains the table name? Parsing the query in this way seems dodgy. De-parsing the COPY statement from the CopyStmt should be straight-forward, since the COPY grammar is fairly simple. We'll take the blame for any changes to the COPY command :).

For comparison, we have a function in CitusDB that de-parses ExplainState and simply includes all the options with their default values:

```
appendStringInfo(explainQuery,
            "EXPLAIN (ANALYZE %s, VERBOSE %s, "
            "COSTS %s, BUFFERS %s, TIMING %s, "
            "FORMAT %s) %s",
            es->analyze ? "TRUE" : "FALSE",
            es->verbose ? "TRUE" : "FALSE",
            es->costs ? "TRUE" : "FALSE",
            es->buffers ? "TRUE" : "FALSE",
            es->timing ? "TRUE" : "FALSE",
            formatStr,
            queryString);
```

COPY is a bit more complicated, but you should be able to take a similar approach with regards to options.

src/pg_copy.c
marcocitus: Seems minor.

```
if (!nextRowFound)
{
    MemoryContextSwitchTo(tupleContext);
    break; 
}
```

src/pg_copy.c
marcocitus: Technically, there can be multiple shards that overlap in which case PruneShardList will return all of them. We would be adding the row to each of them and thus it would show up multiple times on SELECT. For simplicity, you could error out if there are multiple shards or add only to the first.

src/pg_copy.c
marcocitus: Makes sense. We mostly want it to be in a separate function to separate concerns. It would be useful to have a generic connection cache API, though that's probably not within the scope of the project.

src/pg_copy.c
marcocitus: You should lock both. The metadata lock prevents concurrent placement additions. LockShardData prevents concurrent updates/deletes and shards moves. When using statement-based replication, updates require an exclusive lock to ensure that they occur in the same order on all replicas. When not using replication, all_modifications_commutative can be turned on and only  master_copy_shard_placement take an exclusive lock.

When grabbing locks on multiple shards, they need to always be acquired in the same order to avoid deadlock. AcquireExecutorShardLocks sorts the list of shards before acquiring the locks.

src/pg_copy.c
marcocitus: Thanks!

src/pg_copy.c
knizhnik: And what is the advantage of this change?

src/pg_copy.c
knizhnik: And how you solving this problem now with INSERT? I do not see any differences with COPY...

src/pg_copy.c
knizhnik: I undestand that we need to lock metadata to protect from changes in shards distribution. But do we need to lock data itself? Isn't it done by PoastgreSQL itself at shards? For example, when I start COPY command on some node, PostgreSQL will lock correspondent table.
Certainly shards needs to be accessed in the same order to avoid deadlocks.
I wonder whether ti will be more efficient to store shard list in metadata already in ascending order so that it will not be necessary to sort them explicitly each tine we get this list?

src/pg_copy.c
marcocitus: Primarily just to move code that prepares/rolls back/commits the transaction on all placements into separate functions to improve readability. Ideally, this could be a utility function that is a standard part of the transaction manager API. 

e.g.:

```
List *shardConnectionList = ListConnections(shardToConnHash);
...
tmgr->PrepareAll(shardConnectionList);
...
tmgr->CommitAll(shardConnectionList);
CloseAll(shardConnectionList);
```

src/pg_shard.c
marcocitus, line 2122: I meant that you could just make a 1 line change to the existing code by replacing this error. PgShardCopy could just assume that the table is distributed and relation != NULL. That would avoid another layer of nesting in PgShardCopy. Moreover, it keeps all the IsDistributedTable logic in this file.

src/pg_copy.c
marcocitus: Preferably not. Worker nodes can suddenly go down due to failures or maintenance. Ingestion should keep working.

Under certain failure scenarios pg_shard prioritizes availability over consistency. For example, when an INSERT succeeds on one placement, but fails on another, the replicas are no longer consistent. In that case, pg_shard marks the placement on which the INSERT failed as inactive such that subsequent queries don't see the inconsistent replica. The user can then use the copy_shard_placement functions (or the proprietary shard-rebalancer) to repair the inactive shard. If the master node fails before updating the shard placement state, pg_shard will still send queries to the placement on which it failed, so in that case there is some inconsistency which can also be resolved using repair functions.

src/pg_shard.c
knizhnik, line 2122: But I still have top copy most of this code to PgShardCopy, because in any case I need "relation", "rawQuery", "tableId" which are extracted here.
Sorry, but I still think that placing all this logic in one place is more important then minimizing changes of existed pg_shard.c code or reducing level of nesting.

marcocitus:
See #156

PR #156, opened 2015-12-24
Add COPY utility to pg_shard

knizhnik:
I fixed problems and implement suggestions mentioned in your comments and answered rest of them. So if there is no reply to the comment, then I have fixed it.


marcocitus:
Thanks for the improvements. It's pretty impressive what you've managed to put together in a few days. I especially like the pluggable transaction manager API. This will also make it much easier to also plug in DTM at some point.

You can use uncrustify to enforce formatting rules. You can install it from http://uncrustify.sourceforge.net/
    uncrustify -c .uncrustify.cfg --no-backup src/copy.c

I'll answer all the open items tomorrow. In general, we would like to see some subproblems moved into separate functions to avoid deep nesting and make it easier to understand the code. We also have a 90 column limit. It would be even better if you can stick to the 80 character limit used by PostgreSQL without wrapping a lot of lines.

src/pg_copy.c
```diff
@@ -0,0 +1,533 @@
...
+} CopyStateData;
+
+/* TODOL should be moved to copy.c */
+static StringInfo CopyGetLineBuf(CopyStateData *cs)
+{
+	return &cs->line_buf;
+}
+
+#define MAX_SHARDS 1001
+#define MAX_REPLICATION_FACTOR 8 /* should be less or equal than 32 */
```
marcocitus: Can we do without this limit? Some users replicate tables to all nodes to allow local joins.

marcocitus:
I've responded to the comments in PR #155 . Will be out for a few days. Happy new year!

src/pg_copy.c
knizhnik: Is it possible to get somewhere replication factor ?
It seems to me that it is specified only at database creation time and is not stored somewhere...
I will be much easier ans faster if I can preallocate array of proper size instead of dynamically reallocating it on demand.

include/connection.h
```diff
@@ -50,10 +49,27 @@ typedef struct NodeConnectionEntry
 } NodeConnectionEntry;
 
 
+
+typedef struct {
+	ShardId shardId;
+	int nReplicas;
```
marcocitus: Minor: our naming convention is "replicaCount"

include/connection.h
```diff
@@ -50,10 +49,27 @@ typedef struct NodeConnectionEntry
...
+	bool* status;
```
marcocitus: Could you use an enum for the status? (e.g. TRANSACTION_OPEN, TRANSACTION_PREPARED).

src/pg_copy.c
```diff
@@ -0,0 +1,552 @@
...
+
+	memset(&info, 0, sizeof(info));
+	info.keysize = sizeof(ShardId);
+	info.entrysize = sizeof(ShardConnections);
+	info.hash = shard_id_hash_fn;
+
+	return hash_create("shardToConn", MAX_SHARDS, &info, HASH_ELEM | HASH_FUNCTION);
+}
+
+static char const* ConstructCopyStatement(CopyStmt *copyStatement, ShardId shardId)
```
marcocitus: minor: we put the function name on a new line. 

Could you run uncrustify on this file? (see PR Comments)

src/pg_copy.c
```diff
@@ -0,0 +1,552 @@
...
+
+static bool PgCopyAbortTransaction(ShardId shardId, PGconn* conn, void* arg, bool status)
+{
+	char const* relationName = (char*)arg;
+	PgShardTransactionManager const* tmgr = &PgShardTransManagerImpl[PgShardCurrTransManager];
+
+	if (status) 
+	{
+		tmgr->RollbackPrepared(conn, relationName, shardId);
+	} else {
```
marcocitus: minor: we put else on a new line

src/pg_copy.c
```diff
@@ -0,0 +1,552 @@
...
+	bool failOK = true;
+	Oid tableId = 0;
+
+	if (relation == NULL)
+	{
+		return false;
+	}
+	tableId = RangeVarGetRelid(relation, NoLock, failOK);
+	if (IsDistributedTable(tableId))
+	{
```
marcocitus: could you move this code block into a separate function? We really need to get this function under 90 columns.

src/pg_copy.c
```diff
@@ -0,0 +1,552 @@
...
+					if (columnNulls[partitionColumn->varattno]) 
+					{ 
+						ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+										errmsg("cannot copy row with NULL value "
+											   "in partition column")));
+					} 
+					rightConst->constvalue = columnValues[partitionColumn->varattno];
+					prunedList = PruneShardList(tableId, whereClauseList, shardIntervalList);
+					
+					foreach(shardIntervalCell, prunedList)
```
marcocitus: Instead of a foreach loop, this should be something like:

```
 shardInterval = (ShardInterval *) linitial(prunedList);
```

We should copy to at most one matching shard.

src/pg_copy.c
```diff
@@ -0,0 +1,552 @@
...
+					{
+						ShardInterval *shardInterval = (ShardInterval *) lfirst(shardIntervalCell);
+						ShardId shardId = shardInterval->id;
+						bool found = false;
+						
+						shardConn = (ShardConnections*)hash_search(shardToConn, &shardInterval->id, HASH_ENTER, &found);
+						if (!found) { 
+							List *finalizedPlacementList = NULL;
+                            int nPlacements = 0;
```
marcocitus: minor: our naming convention is placementCount

src/pg_copy.c
```diff
@@ -0,0 +1,552 @@
...
+							shardConn->status = (bool*)palloc0(nPlacements*sizeof(bool));
+                            nPlacements = 0;
+
+							foreach(taskPlacementCell, finalizedPlacementList)
+							{
+								ShardPlacement *taskPlacement = (ShardPlacement *) lfirst(taskPlacementCell);
+								char *nodeName = taskPlacement->nodeName;
+								char nodePort[16];
+								PGconn* conn = NULL;
+								sprintf(nodePort, "%d", taskPlacement->nodePort);
```
marcocitus: Hmz, I guess the ConnectToNode API kind of sucks for requiring you to use a string for the port number. Would you mind adding a small wrapper function that does the conversion instead of exposing ConnectToNode directly?

src/pg_copy.c
```diff
@@ -0,0 +1,552 @@
...
+										!PgShardExecute(conn, PGRES_COPY_IN, copy))
+									{
+										elog(ERROR, "Failed to start '%s' on node %s:%s", copy, nodeName, nodePort);
+									}
+								} else { 
+									elog(ERROR, "Failed to connect to node %s:%s", nodeName, nodePort);
+								}
+							}
+						}
+						/* TODO: handle binary format */
```
marcocitus: I guess we can remove these TODO's since it's obvious that we don't support it.

src/pg_copy.c
```diff
@@ -0,0 +1,552 @@
...
+						lineBuf = CopyGetLineBuf(copyState);
+						lineBuf->data[lineBuf->len++] = '\n'; /* there was already new line in the buffer, but it was truncated: no need to heck available space */
+						/* Replicate row to all shards */
```
marcocitus: comment should probably say "shard placements"

src/pg_copy.c
```diff
@@ -0,0 +1,552 @@
...
+						for (i = 0; i < shardConn->nReplicas; i++) { 
+							PQputCopyData(shardConn->conn[i], lineBuf->data, lineBuf->len);
```
marcocitus: What if there's an error here? Does it always fall into PG_CATCH()? The documentation says it just returns -1.

src/pg_copy.c
```diff
@@ -0,0 +1,552 @@
...
+			}
+
+			while (nextRowFound)
+			{
+				MemoryContext oldContext = MemoryContextSwitchTo(tupleContext);
+				nextRowFound = NextCopyFrom(copyState, NULL, columnValues, columnNulls, NULL);
+				MemoryContextSwitchTo(oldContext);
+					
+				/* write the row to the shard */
+				if (nextRowFound)
```
marcocitus: again, to help reduce nesting and bring this function under 90 columns (with 4-space tabs). You could do the following:

```
while (true)
{
    ...
    if (!nextRowFound)
    {
        MemoryContextReset(tupleContext);
        break;
    }
    ...
    MemoryContextReset(tupleContext);
}
```

src/pg_copy.c
```diff
@@ -0,0 +1,552 @@
...
+					prunedList = PruneShardList(tableId, whereClauseList, shardIntervalList);
+					
+					foreach(shardIntervalCell, prunedList)
+					{
+						ShardInterval *shardInterval = (ShardInterval *) lfirst(shardIntervalCell);
+						ShardId shardId = shardInterval->id;
+						bool found = false;
+						
+						shardConn = (ShardConnections*)hash_search(shardToConn, &shardInterval->id, HASH_ENTER, &found);
+						if (!found) {
```
marcocitus: Could you move this code block in to a function like:

```
static int
InitializeShardConnections(ShardConnections *shardConnections,
                           ShardId shardId,
                           PgShardTransactionManager *transactionManager) 
```

(plus whatever parameters I'm forgetting here)

include/connection.h
```diff
@@ -50,10 +49,27 @@ typedef struct NodeConnectionEntry
...
+extern PGconn* ConnectToNode(char *nodeName, char *nodePort);
+
+typedef bool (*ShardAction)(ShardId id, PGconn* conn, void* arg, bool status);
 
+/* 
+ * Perform action for all shards and all shard replicas.
+ * Returns number of shard for which operation od failed or INVALID_SHARD_ID in case of success
+ */
+struct HTAB;
+extern ShardId DoForAllShards(struct HTAB* shardConnHash, ShardAction action, void* arg);
```
marcocitus: I'd prefer not to require all callers to construct an HTAB, could this be changed to a List *shardConnList? 

(the COPY implementation will have the convert the hash table to a list, but this is a very minor overhead)

include/connection.h
knizhnik: My intention was to leave semantic of "status" field for users of DoForAllShards function - for callbacks passed in this function. So there are two options:
1. Make it boolean with semantic: ok/error
2. Make it integer and let callback interpret it any way it like. But in this case ShardAction should also return integer, not boolean. But then return code can not be interpreted by DoForAllShards to stop iteration as soon as possible when error is encountered and return ID of failed shard.

This is why I have chosen option 1.

src/pg_copy.c
knizhnik: Are you sure? It seems to me that in case of data replication (replication factor > 1) we should copy data to all replicas, do we?

src/pg_copy.c
```diff
@@ -0,0 +1,574 @@
...
+			if (columnNulls[partitionColumn->varattno]) 
+			{ 
+				ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+								errmsg("cannot copy row with NULL value "
+									   "in partition column")));
+			} 
+			rightConst->constvalue = columnValues[partitionColumn->varattno];
+			prunedList = PruneShardList(tableId, whereClauseList, shardIntervalList);
+				
+			foreach(shardIntervalCell, prunedList)
```
marcocitus: A shardInterval represents a single shard (part of the table). If you copy into multiple shards, you're copying into the table multiple times. 

The reason PruneShardList might return multiple shards is that a) SELECT/UPDATE/DELETE queries could hit multiple shards, b) multiple shards could have overlapping intervals if users manually change the metadata. We only need to consider case b here. If multiple shards overlap, we should copy to one of them. For correctness, it does not matter which one, so you can just use linitial(prunedList) and remove the foreach loop. Copying to all replicas of the shard is handled at line 539.

include/pg_shard.h
```diff
@@ -77,6 +77,9 @@ typedef struct DistributedPlan
 } DistributedPlan;
 
 
+#define INVALID_SHARD_ID (-1)
+typedef int64 ShardId;
```
marcocitus, line 81: Was there a strong need for this change? I think it looks cleaner, but it will make it harder to integrate.

src/pg_tmgr.c
```diff
@@ -0,0 +1,110 @@
...
+{
+	return true;
+}
+
+/* 
+ * Two-phase commit 
+ */ 
+static char* PgShard2pcCommand(char const* cmd, char const* relationName, int64 shardId)
+{
+    return psprintf("%s 'pgshard_%s_%d'", cmd, relationName, (int)shardId);
```
marcocitus: What if there are 2 concurrent copy transactions, won't they use the same prepared transaction name?

include/pg_tmgr.h
```diff
@@ -0,0 +1,28 @@
...
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#ifndef PG_TMGR_H
+#define PG_TMGR_H
+
+typedef struct { 
+	bool (*Begin)(PGconn* conn);
+	bool (*Prepare)(PGconn* conn, char const* relationName, ShardId shardId);
```
marcocitus: It might be worth calling "relationName" and "shardId" something like "prefix" and "transactionId". Since they are only used to give the transaction a name, they don't need to an actual relation name or shard id and we might want to use these functions for other purposes like multi-shard updates.

include/connection.h
```diff
@@ -50,10 +49,26 @@ typedef struct NodeConnectionEntry
 } NodeConnectionEntry;
 
 
+
+typedef struct {
+	ShardId shardId;
+	int replicaCount;
+	bool* status;
```
marcocitus: if we're keeping it as a bool, can we rename it to "sentPrepare"? That would make its purpose more clear (also applies to functions where its passed as a parameter).

src/connection.c
```diff
@@ -330,3 +328,24 @@ ConnectionGetOptionValue(PGconn *connection, char *optionKeyword)
...
+
+	foreach (listCell, shardConnections)
+	{
+		ShardConnections* shardConn = (ShardConnections*)lfirst(listCell);
+		bool allOk = true;
+		for (i = 0; i < shardConn->replicaCount; i++) 
+		{						
+			allOk &= shardConn->status[i] = action(shardConn->shardId, shardConn->conn[i], arg, shardConn->status[i]);
+		}
+		if (!allOk)
```
marcocitus: DoForAllShards terminates after a single shard fails to perform commit/abort, but this would leave a lot of prepared transactions for all the other shards. I think it would be preferable to produce a warning (e.g. "WARNING: transaction data_192 failed to commit on 192.168.2.3:5432") and continue committing/aborting the other ones, such that the user can step in and recover only those transactions that failed. When a prepare fails we do want to stop immediately and roll back. Having a generic DoForAllShards might not be the right answer here.

Given that the loop is pretty simple now, you could move it into the CopyPrepareTransaction, CopyAbortTransaction, and CopyEndTransaction functions instead of having those act on a single shard and produce the appropriate warnings there.

src/pg_tmgr.c
```diff
@@ -0,0 +1,110 @@
...
+PgShardTransactionManager const PgShardTransManagerImpl[] = 
+{
+	{ PgShardBeginStub, PgShardPrepareStub, PgShardCommitPreparedStub, PgShardRollbackPreparedStub, PgShardRollbackStub },
+	{ PgShardBegin2PC, PgShardPrepare2PC, PgShardCommitPrepared2PC, PgShardRollbackPrepared2PC, PgShardRollback2PC }
+};
+
+/* 
+ * Transaction manager stub
+ */
+static bool PgShardBeginStub(PGconn* conn)
```
marcocitus: For all functions: the name of the function should be on a new line (like postgres source code)

src/pg_copy.c
```diff
@@ -0,0 +1,574 @@
...
+			char const* copy = ConstructCopyStatement(copyStatement, shardId);
+			shardConnections->conn[placementCount++] = conn;
+			/*
+			 * New connection: start transaction with copy command on it.
+			 * Append shard id to table name.
+			 */
+			if (!transactionManager->Begin(conn) ||
+				!PgShardExecute(conn, PGRES_COPY_IN, copy))
+			{
+				elog(ERROR, "Failed to start '%s' on node %s:%d", copy, nodeName, taskPlacement->nodePort);
```
marcocitus: The way this works for insert/update/delete is that we mark the shard placement as inactive on failure and continue with the remaining shard placements (unless all fail). When there are many worker nodes, the chances of one of them being down at any given time are quite high, so we prefer if commands can still go through at the cost of having to repair shards later on. Right now, COPY would fail if a single worker node is down (and has active placements), which would not be desirable. You can look at ExecuteDistributedModify for an example of how to update the shard placements.

src/pg_copy.c
```diff
@@ -0,0 +1,574 @@
...
+					{
+						elog(ERROR, "Copy failed for shard %ld", (long)shardId);
+					}
+				}
+			}
+			MemoryContextReset(tupleContext);
+		}
+		/* Perform two phase commit in replicas */
+		shardConnectionsList = HTABToList(shardToConn);
+		failedShard = DoForAllShards(shardConnectionsList, PgCopyPrepareTransaction, relationName);
```
marcocitus: if we failed to prepare a shard, shouldn't we throw an error and roll back here?

src/pg_copy.c
```diff
@@ -0,0 +1,574 @@
...
+		DoForAllShards(shardConnectionsList, PgCopyAbortTransaction, relationName);
+		PG_RE_THROW();
+	}
+	PG_END_TRY();
+
+	EndCopyFrom(copyState);
+	heap_close(rel, AccessShareLock);
+
+	/* Complete two phase commit */
+	DoForAllShards(shardConnectionsList, PgCopyEndTransaction, relationName);
```
marcocitus: If I understand correctly, we would still commit the prepares that succeeded even if one of them failed right now.

src/pg_shard.c
```diff
@@ -209,6 +216,12 @@ _PG_init(void)
 							 &LogDistributedStatements, false, PGC_USERSET, 0, NULL,
 							 NULL, NULL);
 
+	DefineCustomEnumVariable("pg_shard.copy_TransManager",
```
marcocitus: Maybe copy_transaction_manager

src/pg_copy.c
```diff
@@ -0,0 +1,574 @@
...
+	PQfinish(conn);
+	return true;
+}
+
+static bool PgCopyEndTransaction(ShardId shardId, PGconn* conn, void* arg, bool status)
+{
+	char const* relationName = (char*)arg;
+	PgShardTransactionManager const* tmgr = &PgShardTransManagerImpl[PgShardCurrTransManager];
+	
+	if (status)
```
marcocitus: I'd expect we only get here if prepare was sent on all connections so status/sentPrepare should never be false. Maybe add an Assert.

src/pg_copy.c
```diff
@@ -0,0 +1,574 @@
...
+	PQputCopyEnd(conn, NULL);
+	return tmgr->Prepare(conn, relationName, shardId);
+}
+
+static bool PgCopyAbortTransaction(ShardId shardId, PGconn* conn, void* arg, bool status)
+{
+	char const* relationName = (char*)arg;
+	PgShardTransactionManager const* tmgr = &PgShardTransManagerImpl[PgShardCurrTransManager];
+
+	if (status)
```
marcocitus: We might be rolling back because a connection died, not sure what the behaviour of this function is in case of a dead connection.

include/pg_shard.h
knizhnik, line 81: It seems to be good idea to use separate type for shardId rahter then use int64 in many places...

include/pg_tmgr.h
knizhnik: Frankly speaking "porefix" means nothing (especially in this context) and transactionId is actually not correct, because transationId is formed from relation name and shardId.
So my interpretation was that this function is given relation-name and shard-id and can perform with them anything it like, for example form transaction-id if it is needed. It is responsibility of this function to construct unique transaction id rather than of caller of this function.

src/connection.c
knizhnik: This is why "status" variable is used.
Please notice that PgCopyAbortTransaction and PgCopyEndTransaction always return true, so in any case all transactions in all connections will be completed and there will be no unclosed prepared transactions.

src/pg_copy.c
knizhnik: So it can happen that some data was copied to one node, then node is crashed, then rest of the data will be copied to another node. As a result we will get inconsistent copy result - some copied data is missed. I do not think that it is good behavior. It seems to be better to return error in this case and repeat COPY.

include/pg_shard.h
marcocitus, line 81: I agree, but we use int64 in many places :). If it's not necessary, could you change it back?

src/pg_copy.c
marcocitus: Why would data be missing on the second node? Isn't the all data written to all the placements?

By marking the placement as failed it is no longer included in the finalized placement list of subsequent selects, thus the data on the node we could not connect to is no longer visible.

src/pg_copy.c
```diff
@@ -0,0 +1,600 @@
...
+	PG_END_TRY();
+
+	EndCopyFrom(copyState);
+	heap_close(rel, AccessShareLock);
+
+	/* Complete two phase commit */
+	if (failedShard != INVALID_SHARD_ID) 
+	{ 
+		DoForAllShards(shardConnectionsList, PgCopyAbortTransaction, NULL);
+		elog(ERROR, "COPY failed for shard %ld", (long)failedShard);
```
marcocitus: This code path appears to be identical to adding:

```
if (failedShard != INVALID_SHARD_ID) 
{
    elog(ERROR, "COPY failed for shard %ld", (long)failedShard);
}
```

before line 574, since the clean-up/abort code will be executed in the PG_CATCH block.

src/pg_copy.c
```diff
@@ -0,0 +1,600 @@
...
+			prunedList = shardListCache[shardHashCode];
+			
+			if (prunedList == NULL)
+			{
+				rightConst->constvalue = partitionColumnValue;
+				prunedList = PruneShardList(tableId, whereClauseList, shardIntervalList);
+				shardListCache[shardHashCode] = prunedList;
+			}
+
+			shardInterval = (ShardInterval *) lfirst(list_head(prunedList));
```
marcocitus: Can use linitial( instead of lfirst(list_head(

marcocitus: At this point, we should check whether hashedValue (for hash-partitioning) or partitionColumnValue (for range-partitioning) actually lies within shardInterval. If not, someone manually changed their metadata and the hash-based lookup didn't work.

src/pg_copy.c
knizhnik: But what if operation failed without throwing exception?

src/pg_copy.c
knizhnik: It seems to eliminate (or at least reduce) benefit of caching...
Why somebody can want to manually modify distribution metadata and break uniform hash function?

src/pg_copy.c
marcocitus: Can you run the following command on new files:

```
uncrustify -c .uncrustify.cfg --no-backup src/pg_copy.c
```

(using the .uncrustify.cfg file in the pg_shard repository)

Note that without --no-backup it will dump the output into a src/pg_copy.c.uncrustify file.

src/pg_copy.c
marcocitus: Not sure if I follow, doesn't the elog throw the exception?

(you should use ereport btw)

src/pg_copy.c
```diff
@@ -0,0 +1,696 @@
...
+	return buf->data;
+}
+
+
+static bool
+PgCopyPrepareTransaction(ShardId shardId, PGconn *conn, void *arg, bool status)
+{
+	PgShardTransactionManager const *tmgr =
+		&PgShardTransManagerImpl[PgShardCurrTransManager];
+	PQputCopyEnd(conn, NULL);
```
marcocitus: According to the PQputCopyEnd docs we should also call PQgetResult to wait for the result.

src/pg_copy.c
```diff
@@ -0,0 +1,696 @@
...
+	return hash_create("shardToConn", MAX_SHARDS, &info, HASH_ELEM | HASH_FUNCTION);
+}
+
+
+static char const *
+ConstructCopyStatement(CopyStmt *copyStatement, ShardId shardId)
+{
+	StringInfo buf = makeStringInfo();
+	ListCell *cell;
+	char sep;
```
marcocitus: these should be initialized (NULL, '\0')

src/pg_copy.c
```diff
@@ -0,0 +1,696 @@
...
+	char const *qualifiedName = quote_qualified_identifier(
+		copyStatement->relation->schemaname,
+		copyStatement->relation->
+		relname);
```
marcocitus: Might be an uncrustify thing, relname seems to fit on the previous line.

src/pg_copy.c
```diff
@@ -0,0 +1,696 @@
...
+			}
+			MemoryContextReset(tupleContext);
+			shardConnections->replicaCount -= errorCount;
+			if (shardConnections->replicaCount == 0) /* if all placements failed, error out */
+			{
+				ereport(ERROR, (errcode(ERRCODE_IO_ERROR),
+								errmsg("Could not copy to any active placements for shard %ld",
+									   (long) shardId)));
+			}
+			else if (errorCount != 0) /* otherwise, mark failed placements as inactive: they're stale */
```
marcocitus: Can we roll everything back by cancelling all the COPY commands? I'm not entirely clear on the error semantics of the libpq copy functions. When does commit happen? If we roll back the placement status doesn't have to change.

I think the COPY failing in-flight is a different case then failing to connect at the start. At this point we've already filtered out dead worker nodes. If something fails, it's either due to an exceptional circumstance or due to a functional failure like a constraint violation. In either case, the old behaviour of rolling back without changing placement status makes sense to me.

src/pg_tmgr.c
```diff
@@ -0,0 +1,134 @@
+#include "postgres.h"
+#include "libpq-fe.h"
+#include "miscadmin.h"
+
+#include "connection.h"
+#include "pg_tmgr.h"
+
+#define MAX_STMT_LEN 1024
```
marcocitus: Seems to be unused.

src/pg_copy.c
```diff
@@ -0,0 +1,696 @@
...
+
+			/* write the row to the shard */
+			if (columnNulls[partitionColumn->varattno])
+			{
+				ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+								errmsg("cannot copy row with NULL value "
+									   "in partition column")));
+			}
+			partitionColumnValue = columnValues[partitionColumn->varattno];
+			hashedValue = DatumGetInt32(FunctionCall1(hashFunction,
```
marcocitus: Could you encapsulate all the shard pruning logic in separate functions/file? Ideally this code would look something like:

```
typedef struct ShardCache
{
    ShardInterval *cache;
    FmgrInfo *hashFunction;
    ... 
};
...
struct ShardCache shardCache = {0};
...
InitializeShardCache(&shardCache);
...
while (true)
{
    ...
    partitionColumnValue = columnValues[partitionColumn->varattno];
    shardInterval = LookupShardByValue(&shardCache, partitionColumnValue);
    ....
}
```

LookupShardByValue should take care of range-partitioning and non-uniform shard distributions as well.

src/pg_tmgr.c
```diff
@@ -0,0 +1,134 @@
...
+PgShardRollbackStub(PGconn *conn)
+{
+	return true;
+}
+
+
+/*
+ * Two-phase commit
+ */
+static int GlobalTransactionId;
```
marcocitus: When is this initialized? Also, global variables should be at the top.

src/pg_copy.c
```diff
@@ -0,0 +1,696 @@
...
+	if (shardIntervalList == NIL)
+	{
+		ereport(ERROR, (errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+						errmsg("could not find any shards for query"),
+						errdetail("No shards exist for distributed table \"%s\".",
+								  relationName),
+						errhint("Run master_create_worker_shards to create shards "
+								"and try again.")));
+	}
+	if (!copyStatement->is_from) /* Construct query selecting data from this relation */
```
marcocitus: I have a slight preference for moving this code block outside of this function to separate concerns. What if we introduce an intermediate function that distinguishes between COPY FROM and COPY TO?

```
void
PgShardCopy(CopyStmt *copyStatement, char const *query)
{
    if (copyStatement->is_from)
    {
        PgShardCopyFrom(copyStatement, query);
    }
    else
    {
        PgShardCopyTo(copyStatement, query);
    }
}
```

PgShardCopyTo would contain the code below.

src/pg_copy.c
```diff
@@ -0,0 +1,696 @@
...
+			placementCount += 1;
+
+			/*
+			 * New connection: start transaction with copy command on it.
+			 * Append shard id to table name.
+			 */
+			if (!transactionManager->Begin(conn) ||
+				!PgShardExecute(conn, PGRES_COPY_IN, copy))
+			{
+				failedPlacementList = lappend(failedPlacementList, taskPlacement);
```
marcocitus: We should probably distinguish between Begin or PgShardExecute failing, and connection failing. Begin could fail due to some issue with the DTM. PgShardExecute could fail because of some issue with the worker node (e.g. shard was deleted). I'm not sure if marking the placements as inactive is the way to go there. Then again, the inactive placements only come in if some of the placements succeed. Anyway, let's discuss this in the meeting :)

knizhnik: There are no reasons for TM (at least for current implementations of TM in pg_tmgr.c) to reject start of new transaction except:
1. Server is in error state
2. Connection failure with node
In all this cases COPY command also can not be started. So I am not sure that we should distinguish this two cases.

test/sql/14-copy-utility.sql
```diff
@@ -0,0 +1,18 @@
...
+SELECT master_create_worker_shards(table_name := 'company',
+                                   shard_count := 16,
+                                   replication_factor := 1);
+\set VERBOSITY default
+\copy company from 'test/data/customer.csv' delimiter ',' csv; 
+select * from company;
+\copy company to 'results.csv';
+\copy (select name from company) to 'names.csv';
+\copy company from 'test/data/format-error.csv' delimiter ',' csv; 
+\copy company from 'test/data/constraint-error.csv' delimiter ',' csv;
```
marcocitus: Might be good to add some COPY tests as well, since this behaves differently from \COPY.

src/connection.c
```diff
@@ -330,3 +328,24 @@ ConnectionGetOptionValue(PGconn *connection, char *optionKeyword)
 
 	return optionValue;
 }
+
+ShardId DoForAllShards(List* shardConnections, ShardAction action, void* arg)
```
marcocitus: We find the failure semantics of the DoForAllShards function a bit confusing. Could you remove it and put the loops in each of the 3 tasks we want to perform?

In other words, you would get the following functions:

```
 static ShardId PgCopyPrepareTransactions(List *shardConnections);
 static ShardId PgCopyAbortTransactions(List *shardConnections);
 static ShardId PgCopyEndTransactions(List *shardConnections);
```

include/pg_tmgr.h
```diff
@@ -0,0 +1,29 @@
+/*-------------------------------------------------------------------------
+ *
+ * include/pg_tmgr.h
```
marcocitus: maybe name it transaction_manager.h and transaction_manager.c, the pg_ prefix suggests this is for PostgreSQL in general.

src/connection.c
```diff
@@ -63,7 +62,7 @@ GetConnection(char *nodeName, int32 nodePort)
 	NodeConnectionEntry *nodeConnectionEntry = NULL;
 	bool entryFound = false;
 	bool needNewConnection = true;
-
+
```
marcocitus: spurious whitespace change

src/pg_copy.c
```diff
@@ -0,0 +1,742 @@
+/*-------------------------------------------------------------------------
+ *
+ * src/pg_copy.c
```
marcocitus: Maybe rename to distributed_copy.c, since the pg_ prefix doesn't mean much.

src/pg_copy.c
```diff
@@ -0,0 +1,742 @@
...
+ *
+ * This file contains implementation of COPY utility for pg_shard
+ *
+ * Copyright (c) 2014-2015, Citus Data, Inc.
+ *
+ ****-------------------------------------------------------------------------
```
marcocitus: Use one asterisk

src/pg_copy.c
```diff
@@ -0,0 +1,742 @@
...
+	memset(&info, 0, sizeof(info));
+	info.keysize = sizeof(ShardId);
+	info.entrysize = sizeof(ShardConnections);
+	info.hash = shard_id_hash_fn;
+
+	return hash_create("shardToConn", MAX_SHARDS, &info, HASH_ELEM | HASH_FUNCTION);
+}
+
+
+static char const *
```
marcocitus: Can you add comments to all non-trivial functions, similar to the one above.

src/pg_copy.c
```diff
@@ -0,0 +1,742 @@
...
+}
+
+
+static bool
+PgCopyAbortTransaction(ShardId shardId, PGconn *conn, void *arg, bool status)
+{
+	PgShardTransactionManager const *tmgr =
+		&PgShardTransManagerImpl[PgShardCurrTransManager];
+
+	if (status)
```
marcocitus: I still don't like the name status, should have a semantic name like sentPrepare.

src/pg_copy.c
marcocitus: Fair point, I was thinking primarily what if the COPY command fails, maybe because a shard was deleted. Then again, marking the placement as inactive is probably better in that case.

test/sql/14-copy-utility.sql
```diff
@@ -0,0 +1,24 @@
...
+                                   shard_count := 16,
+                                   replication_factor := 1);
+\set VERBOSITY default
+\copy company from 'test/data/customer.csv' delimiter ',' csv; 
+select * from company;
+\copy company to 'results.csv';
+\copy (select name from company) to 'names.csv';
+\copy company from 'test/data/format-error.csv' delimiter ',' csv; 
+\copy company from 'test/data/constraint-error.csv' delimiter ',' csv; 
+copy company from stdin delimiter ';' null '???';
```
marcocitus: In particular, I'd be interested in COPY company FROM 'test/data/constraint-error.csv', since this behaves differently from \COPY company FROM 'test/data/constraint-error.csv'. Also COPY company FROM PROGRAM ...

src/connection.c
knizhnik: You really think that duplicating the same code (two nested loops) in three places is really good idea?
What is confusing with DoForAllShards? It is standard map function...

include/pg_tmgr.h
knizhnik: pg_shard.c, pg_copy.c, ... ?

test/sql/14-copy-utility.sql
knizhnik: I do not know portable way of specifying path to the file for command executed at server.
Postgres COPY test doesn't contain such cases.

src/pg_copy.c
```diff
@@ -0,0 +1,742 @@
...
+									   "in partition column")));
+			}
+			partitionColumnValue = columnValues[partitionColumn->varattno-1];
+			hashedValue = DatumGetInt32(FunctionCall1(hashFunction,
+													  partitionColumnValue));
+			shardHashCode =
+				(int) ((uint32) (hashedValue - INT32_MIN) / hashTokenIncrement);
+			prunedList = shardListCache[shardHashCode];
+
+			if (prunedList == NULL)
```
marcocitus: With the current code, it seems it's possible that we get a cache hit (prunedList != NULL) on something that actually falls outside the cached shard interval. 

For the cache, you might just want to cache the first shard interval instead of the whole list.

include/pg_tmgr.h
marcocitus: pg_shard stands for postgres sharding and is the name of the extension (like pg_stat_statements, pg_partman, etc.). This file provides transaction management function specific to pg_shard, not postgres in general. I guess pgs_tmgr.c would be more accurate, but we usually prefer words over acronyms.

src/pg_copy.c
knizhnik: Can you explain more precisely how it can happen if distributed metadata can not be changed during copy processing (and cache lifetime is limited to COPY)

src/connection.c
marcocitus: Yes. It's reasonably simple code, so the redundancy from having it multiple times is small. My concern is that the failure semantics are different depending on how this function is called. If a prepare fails we should stop immediately, and if an abort fails we should keep going. It's not clear from the DoForAllShards function that it does this to an uninformed reader, since he wouldn't don't know the return values of the functions, nor is it clear from looking at the call to DoForAllShards, nor is it clear from looking at PgCopyPrepareTransaction or PgCopyEndTransaction individually. He will have to look at 3-5 different places and think about how they tie together before knowing what happens in case of failure. By making the behaviour explicit in separate functions the behaviour will be obvious and will only take a few seconds to understand, which makes it much easier to maintain the code even if it is a few more lines.

src/pg_copy.c
```diff
@@ -0,0 +1,769 @@
...
+ */
+static bool
+PgCopyAbortTransaction(ShardId shardId, PGconn *conn, void *arg, bool isPrepared)
+{
+	PgShardTransactionManager const *tmgr =
+		&PgShardTransManagerImpl[PgShardCurrTransManager];
+
+	if (isPrepared)
+	{
+		tmgr->RollbackPrepared(conn);
```
marcocitus: If a RollbackPrepared fails, can we produce some explicit warning message that will help the user to roll it back manually?

src/pg_copy.c
```diff
@@ -0,0 +1,769 @@
...
+ * This function is called only of COPY is successfully completed at all nodes
+ */
+static bool
+PgCopyEndTransaction(ShardId shardId, PGconn *conn, void *arg, bool isPrepared)
+{
+	PgShardTransactionManager const *tmgr =
+		&PgShardTransManagerImpl[PgShardCurrTransManager];
+
+	Assert(isPrepared);
+	tmgr->CommitPrepared(conn);
```
marcocitus: If a CommitPrepared fails, can we produce some explicit warning message that will help the user to commit manually?

test/sql/14-copy-utility.sql
marcocitus: The regression test framework allows you to create an input/ and output/ folder in the regression tests, which contain templates for the sql/ and expected/ files which can contain the source folder as a variable.

e.g. in src/test/regress/input/copy.source

    COPY testjsonb FROM '@abs_srcdir@/data/jsonb.data';

test/sql/14-copy-utility.sql
```diff
@@ -0,0 +1,24 @@
...
+C108;Siemems
+C109;???
+C110;IBM
+\.
+select * from company;
```
marcocitus, line 25: Might be interesting to use: COPY company TO stdout  here, to cover that case as well.

src/pg_copy.c
marcocitus: There are many partitionColumnValues that map to the same shardHashCode. The first row that maps to a shardHashCode might match the interval and cause the interval to be cached for that shardHashCode, but the second row that maps to the same shardHashCode might not fall within the interval.

src/connection.c
knizhnik: According to this logic everything should be placed in one function... Only in this case it should not be necessary to look into several places. 

DoForAllShards implements abstraction of iteration. So if programmer need to understand how shard placements are traversed, he/she should look only at this function. PgCopy*Transaction are operations performed with this shard placements. It has no relation with traversal rule.
Still do not think that elimination of generic iteration function is good idea.

src/distributed_copy.c
```diff
@@ -0,0 +1,857 @@
...
+						break;
+					}
+					else
+					{
+						l = m + 1;
+					}
+				}
+			}
+			if (shardInterval == NULL)
+			{
```
marcocitus: I think we should now be able to find any existing shard in the cache, so if we don't we should just error out:

You can use the error message from https://github.com/citusdata/pg_shard/blob/develop/src/pg_shard.c#L738

src/distributed_copy.c
```diff
@@ -0,0 +1,857 @@
...
+				int32 shardMinHashToken = INT32_MIN + (shardId * hashTokenIncrement);
+				int32 shardMaxHashToken = shardMinHashToken + (hashTokenIncrement - 1);
+				if (shardId == (shardCount - 1))
+				{
+					shardMaxHashToken = INT32_MAX;
+				}
+				if (DatumGetInt32(shardInterval->minValue) != shardMinHashToken ||
+					DatumGetInt32(shardInterval->maxValue) != shardMaxHashToken)
+				{
+					partitionType = RANGE_PARTITION_TYPE;
```
marcocitus: I think the technique of checking the intervals in advance and then falling back to binary search is pretty neat. It might make sense to introduce a separate variable (like searchAlgorithm = BINARY_SEARCH), because you still need to calculate the hash value if the table is hash-partitioned and search for that.

src/distributed_copy.c
```diff
@@ -0,0 +1,857 @@
...
+	List *queryList = raw_parser(psprintf("select * from %s", qualifiedName));
+	copyStatement->query = linitial(queryList);
+	copyStatement->relation = NULL;
+	
+	/* Collecting data from shards will be done by select handler */
+	DoCopy(copyStatement, query, &processedCount);
+}	
+
+static int
+CompareTaskByMinHashToken(const void *key1, const void *key2, void *arg)
```
marcocitus: This function should probably be named CompareShardIntervalsByMinValue.

src/distributed_copy.c
```diff
@@ -0,0 +1,857 @@
...
+{
+	Datum		d1 = (*(ShardInterval**)key1)->minValue;
```
marcocitus: should use descriptive variable names (maybe leftShardInterval, rightShardInterval, comparisonResult)

src/distributed_copy.c
```diff
@@ -0,0 +1,857 @@
...
+/*
+ * Send PQputCopyEnd command to the client and check result (status of 
+ * COPY command completion)
+ */
+static bool
+PgCopyEnd(PGconn *con, char const* msg)
+{
+	PGresult *result = NULL;
+	if (PQputCopyEnd(con, msg) == 1)
+	{
```
marcocitus, line 305: for readability, maybe just return early if there is an error

```
if (PQputCopyEnd(connection, msg) != 1)
{
    /* Connection failure */
    ReportRemoteError(connection, NULL);
    return false;
}

while ...
```

src/distributed_copy.c
```diff
@@ -0,0 +1,857 @@
...
+		{
+			PGconn* conn = shardConn->placements[i].conn;
+			shardConn->placements[i].copied = true;
+			if (PgCopyEnd(conn, NULL) && tmgr->Prepare(conn))
+			{
+				shardConn->placements[i].prepared = true;
+			}
+			else
+			{
+				return shardConn->shardId;
```
marcocitus, line 352: I think it would be nicer to just do an ereport(ERROR here and return void, effectively moving the message from the end of PgShardCopyFrom to here.

src/distributed_copy.c
```diff
@@ -0,0 +1,857 @@
...
+
+/* TODOL should be moved to copy.c */
+static StringInfo
+CopyGetLineBuf(CopyStateData *cs)
+{
+	return &cs->line_buf;
+}
+
+
+#define MAX_SHARDS 1001
```
marcocitus: This should probably be named more appropriately, maybe INITIAL_CONNECTION_CACHE_SIZE and moved to the top below the #includes. You could also consider using MaxConnections instead.

src/distributed_copy.c
knizhnik: Yes, but in case of hash partitioning it can be null

src/distributed_copy.c
knizhnik: Sorry, I do not understand your argument. If I find out that distribution is manually changed by user and so standard hash function can not be used, then I change partitionType to RANGE_PARTITION_TYPE. In this case hash is not calculated. Do not see much reasons to introduce extra variable and enum.

src/distributed_copy.c
knizhnik: I just want to make it "compatible" with CompareTasksByShardId which also manipulates with ShardIntervals.

src/distributed_copy.c
knizhnik, line 305: Ok, but IMHO it reduces readability

src/distributed_copy.c
marcocitus: If a table is hash-partitioned, then minValue and maxValue represent a range of hash values. The ranges can be modified, in which case doing binary search makes a lot of sense, but they still represent ranges of hash values, so we need to calculate the hash of the partition column value (which can be any type) and find the corresponding range of integers.

knizhnik: O, I see...

src/distributed_copy.c
```diff
@@ -0,0 +1,857 @@
...
+	List *queryList = raw_parser(psprintf("select * from %s", qualifiedName));
+	copyStatement->query = linitial(queryList);
+	copyStatement->relation = NULL;
+	
+	/* Collecting data from shards will be done by select handler */
+	DoCopy(copyStatement, query, &processedCount);
+}	
+
+static int
+CompareShardIntervalsByMinHashToken(const void *leftElement,
```
marcocitus: My main objection is actually against HashToken, which should be Value I think, since we're comparing the minValue's.

src/distributed_copy.c
marcocitus: Ow, right. In my head was assuming that we were populating the cache at the start now. Would that be possible in the hash-partitioned case as well?

src/distributed_copy.c
```diff
@@ -0,0 +1,857 @@
...
+
+		foreach(shardIntervalCell, shardIntervalList)
+		{
+			ShardInterval *shardInterval = (ShardInterval *) lfirst(shardIntervalCell);
+			ShardId shardId = shardInterval->id;
+			LockShardData(shardId, ShareLock);
+			LockShardDistributionMetadata(shardId, ShareLock);
+			if (partitionType == HASH_PARTITION_TYPE)
+			{
+				int32 shardMinHashToken = INT32_MIN + (shardId * hashTokenIncrement);
```
marcocitus: shardId should probably be something else (shardIndex?). The master_create_worker_shards worker functions starts counting shardIndex from 0, while shardId could be anything.

src/distributed_copy.c
knizhnik: Sorry, I do not understand to which code fragment this comment is related

expected/copy-utility.source
```diff
@@ -0,0 +1,58 @@
+CREATE TABLE customer
```
marcocitus, line 1: I think this file is stale / superseded by input/copy-utility.source

src/distributed_copy.c
marcocitus: The name of the function contains HashToken, but this should be Value since the value may not be a hash token.

s/CompareShardIntervalsByMinHashToken/CompareShardIntervalsByMinValue/

src/distributed_copy.c
```diff
@@ -0,0 +1,845 @@
...
+				}
+			}
+			if (!useBinarySearch)
+			{
+				shardHashCode =
+					(int) ((uint32) (hashedValue - INT32_MIN) / hashTokenIncrement);
+				shardInterval = shardIntervalCache[shardHashCode];
+			}
+			else
+			{
```
marcocitus, line 805: Can we move this block into a separate function, e.g.

```
shardInterval = FindShardIntervalInCache(shardIntervalCache, partitionColumnValue, compareFunction);
```

Also, use descriptive variable names, maybe l -> lowerBoundIndex, r -> upperBoundIndex, m -> middleIndex

test/sql/14-copy-utility.sql
```diff
@@ -0,0 +1,26 @@
+CREATE TABLE company
```
marcocitus, line 1: This file also seems stale / superseded by test/output/copy-utility.source.

test/expected/10-utilities.out
```diff
@@ -10,11 +10,17 @@ SELECT master_create_distributed_table('sharded_table', 'id');
 
 -- COPY is not supported with distributed tables
```
marcocitus: Maybe just remove these tests since COPY is actually supported now, although it might be useful to add a test to copy-utility.source where we try to COPY into a distributed table without shards.

expected/copy-utility.source
knizhnik, line 1: This file was added just  to make directory non-empty, otherwise I failed to put in repository and without it regression test doesn't work,

src/distributed_copy.c
knizhnik, line 805: Do not think that it is good idea.
We need to pass the following parameters to this function:
FindShardIntervalInCache(partitionType, useBinarySearch, shardIntervalCache, shardCount, partitionColumnValue, compareFunction);
and recalculate hashTokenIncrement (extra division).

Concerning l -> lowerBoundIndex, r -> upperBoundIndex, m -> middleIndex - in all books where I see binary search algorithm, for example in Wirth's "Algorithms+Structures=Programs", exactly these variables were used:)

test/sql/14-copy-utility.sql
knizhnik, line 1: This file is testing \copy directive while  test/output/copy-utility.source tests COPY SQL command (although I certainly understand that \copy is tranformed into COPY FROM STDIN).

marcocitus, line 1: ah, ok, that's fine.

src/distributed_copy.c
marcocitus, line 805: I agree, I actually only meant the binary search part, which should only have those 3 parameters.

marcocitus:
Thanks! I'm running some tests and we're doing an internal demo today

A few open issues.

Regression tests fail on travis (see bottom of PR page) because of different port numbers.

Pressing Ctrl+C while running:

```
COPY customer_reviews FROM '/home/ec2-user/customer_reviews_1998.csv' WITH CSV;
```

causes a segmentation fault

Pressing Ctrl+C while running:

```
\COPY customer_reviews FROM 'customer_reviews_1998.csv' WITH CSV;
```

doesn't roll back the transaction (should check postgres' QueryCancelPending)

The number of rows is missing from the output when copying to a distributed table:

```
postgres# \COPY customer_reviews_local FROM 'customer_reviews_1998.csv' WITH CSV;
COPY 589859
postgres# \COPY customer_reviews FROM 'customer_reviews_1998.csv' WITH CSV;
COPY
```

marcocitus:
When I try to use the 2PC transaction manager, I get a lot the following result:

```
postgres=# SET pg_shard.copy_transaction_manager TO '2pc';
SET
postgres=# \COPY customer_reviews FROM 'customer_reviews_1998.csv' WITH CSV;
WARNING:  Bad result from ip-10-186-176-41.ec2.internal:5432
DETAIL:  Remote message: prepared transaction with identifier "pgshard_28544_256" does not exist
WARNING:  Failed to commit prepared transaction for shard 10127
....
```

It appears to be using the same transaction id for all transactions when committing.

marcocitus:
In PostgreSQL's COPY, we get to see the line number when there is a parsing error. 

```
postgres=# \COPY customer_reviews FROM 'invalid-customer_reviews_1998.csv' WITH CSV;
ERROR:  missing data for column "review_rating"
postgres=# \COPY customer_reviews_local FROM 'invalid-customer_reviews_1998.csv' WITH CSV;
ERROR:  missing data for column "review_rating"
CONTEXT:  COPY customer_reviews_local, line 589860: "ATVPDKIKX0DER,"
```

Could we get the same in distributed COPY?

marcocitus:
I noticed that a regular user can perform COPY from/to a file or program. In PostgreSQL's copy.c there are some check to ensure only superusers can perform these. Could you add those? Ideally, we would check permissions as well.

marcocitus:
I now get a segmentation fault if I cancel twice:

```
postgres=# \COPY customer_reviews FROM 'customer_reviews_1998.csv' WITH CSV;
^CCancel request sent
ERROR:  COPY was canceled
Time: 5721.356 ms
postgres=# \COPY customer_reviews FROM 'customer_reviews_1998.csv' WITH CSV;
^CCancel request sent
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
Time: 671.860 ms
```

knizhnik:
> In PostgreSQL's COPY, we get to see the line number when there is a parsing error. 
> 1. I think that will be necessary to transfer message detail.
> Right now ReportRemoteError is not extracting PG_DIAG_MESSAGE_DETAIL and use primary message as message detail:
>  errdetail("Remote message: %s", remoteMessage)
> Do you want me to change it?

knizhnik:
Another issue is that reported line will be valid only for data redirected to this shard. There is not mapping with lines in the original file and hard to reconstruct this mapping.
So may be there is no much sense in trying to copy detailed error message from replica?

marcocitus:
> 1. I think that will be necessary to transfer message detail.

The worker node doesn't know what the original line number in the file was. I think it would be more useful to count the line number on the master and show it in a message in the PG_CATCH block.

knizhnik:
> The worker node doesn't know what the original line number in the file was. I think it would be more useful to count the line number on the master and show it in a message in the PG_CATCH block.

Please find that error happens on replica, not at mast and due to asynchronous nature of PQputCopyData/PQputCopyEnd we do not know when precisely it happen

knizhnik:
Can not reproduce the crash:
postgres=# copy company from '/tmp/company.csv'  WITH CSV;
^CCancel request sent
ERROR:  COPY was canceled
postgres=# copy company from '/tmp/company.csv'  WITH CSV;
^CCancel request sent
ERROR:  COPY was canceled
postgres=# copy company from '/tmp/company.csv'  WITH CSV;
COPY 1000000
postgres=# ^C

knizhnik:
postgres=# \copy company from '/tmp/company.csv'  WITH CSV;
^CERROR:  COPY from stdin failed: canceled by user
postgres=# \copy company from '/tmp/company.csv'  WITH CSV;
^CCancel request sent
ERROR:  COPY was canceled
postgres=# \copy company from '/tmp/company.csv'  WITH CSV;
^CCancel request sent
ERROR:  COPY was canceled
postgres=# \copy company from '/tmp/company.csv'  WITH CSV;
COPY 1000000

marcocitus:
> Please find that error happens on replica, not at mast and due to asynchronous nature of PQputCopyData/PQputCopyEnd we do not know when precisely it happen

For parsing errors, doesn't NextCopyFrom error out?

knizhnik:
> I noticed that a regular user can perform COPY from/to a file or program. In PostgreSQL's copy.c there are some check to ensure only superusers can perform these. Could you add those? Ideally, we would check permissions as well.

Permissions will be in any case checked at replicas. Why do we need to do it at master?

marcocitus:
> Permissions will be in any case checked at replicas. Why do we need to do it at master?

When you do a COPY FROM 'file' that file is read on the master, and if you do COPY FROM PROGRAM, that program is executed on the master. Only superuser should be allowed to do this. The workers just get a COPY FROM STDIN, which doesn't require superuser.

marcocitus:
Looks pretty good now. We're happy about the performance and behaviour.

I am seeing some issues when something fails during the COPY. If I do killall postgres on a worker node while a COPY is going on, then the master gets a segmentation fault.

I'm also seeing segmentation faults when trying to ingest many files in parallel (exhausting max_connections on the workers). I was running:

```
split -nl/64 customer_reviews_1998.csv chunks/
find chunks -type f | time xargs -t -n1 -P64 sh -c "psql -c \"\\COPY customer_reviews FROM '\$0' WITH CSV\""
```

knizhnik:
Did you use the latest version? Yesterday I have fixed one reason of SIGSEGV.
I can not reproduce the error:
knizhnik@knizhnik:~/dtm-data/customer_reviews$ find chunks -type f | time xargs -t -n1 -P64 sh -c "psql postgres -c \"\COPY customer_reviews FROM '\$0' WITH CSV\""
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ba 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ad 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/az 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/an 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/cj 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bu 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/cb 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ax 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bf 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bp 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bx 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ak 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/af 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/cd 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ao 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/be 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ay 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bl 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/cc 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bo 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/cg 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/am 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/av 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bi 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ab 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ca 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bg 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/aj 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bk 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bc 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ag 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ah 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/as 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bj 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bb 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/cl 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bv 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bz 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/aa 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bw 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/cf 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/al 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/au 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bh 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bm 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/aw 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/aq 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/br 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bs 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ck 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ae 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ar 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ci 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bd 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/at 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bq 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ch 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bt 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/by 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/bn 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ap 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ac 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ai 
sh -c psql postgres -c "\COPY customer_reviews FROM '$0' WITH CSV" chunks/ce 
COPY 18286
COPY 18267
COPY 18316
COPY 18271
COPY 18306
COPY 18212
COPY 18290
COPY 18356
COPY 18323
COPY 18434
COPY 18367
COPY 18302
COPY 18410
COPY 18293
COPY 18145
COPY 18252
COPY 18245
COPY 18202
COPY 18341
COPY 18339
COPY 18379
COPY 18207
COPY 18399
COPY 18408
COPY 18387
COPY 18265
COPY 18347
COPY 18360
COPY 18421
COPY 18341
COPY 18272
COPY 18246
COPY 18415
COPY 18228
COPY 18488
COPY 18370
COPY 18369
COPY 18382
COPY 18377
COPY 18322
COPY 18394
COPY 18353
COPY 18262
COPY 18348
COPY 18245
COPY 18247
COPY 18400
COPY 18194
COPY 18349
COPY 18332
COPY 18378
COPY 18397
COPY 18225
COPY 18362
COPY 18229
COPY 18441
COPY 18389
COPY 18360
COPY 18347
COPY 18327
COPY 18168
COPY 18295
COPY 18245
COPY 18418
0.29user 0.24system 0:04.31elapsed 12%CPU (0avgtext+0avgdata 3644maxresident)k
0inputs+0outputs (0major+17163minor)pagefaults 0swaps

marcocitus:
You're right, that fixes it. Great!

PR #160, opened 2016-02-10
Fix parallel loading command in README

marcocitus:
ls prints file names without paths, which means you'd either have to cd into the chunks directory first or use find.


onderkalaci:
@jasonmp85 @marcocitus Do you think should we check this in? We had a problem in the parallel loading, and I realized this PR, which helped us to solve.

